PR #73, opened 2019-08-12
Peer factory

alicevhenshaw:
### Description

- A factory to allow trusted Peer contracts to be deployed - the factory is IWhitelist (see below)
- Locators back to the more generic bytes32 to enable IPFS hashes as peers in future.
- Indexer can take an IWhitelist to check whether a peer is 'trusted'
- Tests are included.

## Coverage

```
------------------|----------|----------|----------|----------|----------------|
File              |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
------------------|----------|----------|----------|----------|----------------|
 contracts/       |      100 |      100 |      100 |      100 |                |
  Consumer.sol    |      100 |      100 |      100 |      100 |                |
  Peer.sol        |      100 |      100 |      100 |      100 |                |
  PeerFactory.sol |      100 |      100 |      100 |      100 |                |
  Market.sol      |      100 |      100 |      100 |      100 |                |
  Indexer.sol     |      100 |      100 |      100 |      100 |                |
------------------|----------|----------|----------|----------|----------------|
------------------|----------|----------|----------|----------|----------------|
```


examples/peer/test/Peer-factory.js
```diff
@@ -0,0 +1,72 @@
...
+        peerOwnerOne
+      )
+      passes(tx)
+
+      // mapping has been updated
+      isTrustedPeer = await peerFactory.trustedPeers.call(peerAddress)
+      isTrue(isTrustedPeer)
+    })
+  })
+})
```
ejwessel: Mind calling this PeerFactoryUnit or something similar to avoid confusion with the integrated test?

examples/peer/migrations/2_deploy_contracts.js
```diff
@@ -2,11 +2,13 @@ const Peer = artifacts.require('Peer')
 const Swap = artifacts.require('@airswap/swap/contracts/Swap.sol')
 const Types = artifacts.require('@airswap/types/contracts/Types.sol')
 
+const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000'
```
ejwessel: can we use: 
`const { EMPTY_ADDRESS } = require('@airswap/order-utils').constants`

examples/peer/contracts/PeerFactory.sol
```diff
@@ -0,0 +1,46 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@airswap/peer/contracts/Peer.sol";
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract PeerFactory {
+
+  event NewTrustedPeer(address peer, address swap, address owner);
+
+  mapping(address => bool) public trustedPeers;
```
ejwessel: why do we need to keep track of `trustedPeers`?

examples/peer/migrations/2_deploy_contracts.js
alicevhenshaw: Not as easily as we can in tests due to the fact that `web3` is not available in `truffle migrate` (whereas it is available in `truffle test`). Seemed easier to just define the constant in here as opposed to importing web3 purely for the purpose of making the constant imports work.

examples/peer/contracts/PeerFactory.sol
alicevhenshaw: The purpose of this factory is so that we can allow people to deploy peers that we know are trustworthy. But by that logic we should therefore have a way of quickly checking if the peers came out of the factory or not.

examples/peer/migrations/2_deploy_contracts.js
ejwessel: > web3 is not available in truffle migrate
🤦‍♂

examples/peer/migrations/2_deploy_contracts.js
```diff
@@ -1,12 +1,16 @@
...
+
 module.exports = deployer => {
   deployer.deploy(Types)
   deployer.link(Types, Swap)
   deployer
     .deploy(Swap)
     .then(() => Swap.deployed())
-    .then(() => deployer.deploy(Peer, Swap.address))
+    .then(() => deployer.deploy(Peer, Swap.address, EMPTY_ADDRESS))
+  deployer.deploy(PeerFactory)
```
ejwessel: can we not do a promise chain and instead deploy line by line like we had been doing in CREV2:https://github.com/airswap/factora-creV2/blob/master/migrations/2_base_migration.js

alicevhenshaw: Yeah sure, I prefer line by line too in migrations. However this is the case in all migrations for consumer, wrapper, peer, indexer and market, so it should be a separate PR/issue to tackle.

examples/peer/contracts/PeerFactory.sol
poiresel: factoryCreatedPeers instead

poiresel: also why not use a mapping of Peers -> bool

alicevhenshaw: Per meeting today the mapping will be more generally named after a whitelist

alicevhenshaw: Will make mapping `IPeer` to `Bool`

examples/peer/test/Peer-factory-unit.js
```diff
@@ -0,0 +1,104 @@
...
+
+    it('should return address, emit event and update the mapping', async () => {
+      // return the address
+      let peerAddress = await peerFactory.deployTrustedPeer.call(
+        swapContractOne,
+        peerOwnerOne
+      )
+
+      // before calling the function, the mapping is false
+      let isTrustedPeer = await peerFactory.isWhitelisted.call(peerAddress)
```
poiresel: add a test where you insert a fully used 32 byte input.

poiresel: per comment in slack will ignore

alicevhenshaw: Will add this once we've agreed on contracts :)

examples/peer/contracts/PeerFactory.sol
```diff
@@ -0,0 +1,56 @@
...
+pragma solidity 0.5.10;
+pragma experimental ABIEncoderV2;
+
+import "@airswap/peer/contracts/Peer.sol";
+import "@airswap/peer/interfaces/IWhitelist.sol";
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract PeerFactory is IWhitelist {
+
+  event NewTrustedPeer(address peer, address swap, address owner);
```
poiresel: NewFactoryPeer also index the addresses

examples/peer/contracts/PeerFactory.sol
alicevhenshaw: Looking at this again, I think that it may make it simpler to leave the mapping as address => bool, so that anyone can easily look up an address in the mapping without having to cast the address to a Peer. Thoughts?

poiresel: i think that would work, given we're changing the purpose in that this function will be called not just for Peers but for external addresses as well.

protocols/indexer/contracts/Indexer.sol
```diff
@@ -39,19 +40,26 @@ contract Indexer is IIndexer, Ownable {
   // Mapping of address to timestamp of blacklisting
   mapping (address => uint256) public blacklist;
 
+  // The whitelist contract for checking whether a peer is whitelisted
+  address whitelist;
```
ejwessel: visibility modifier missing.

examples/peer/contracts/PeerFactory.sol
```diff
@@ -0,0 +1,56 @@
...
+  function deployTrustedPeer(address _initialSwapContract, address _peerOwner) public returns (address) {
+    require(_peerOwner != address(0), 'Provide a peer owner');
+    require(_initialSwapContract != address(0), 'Provide a swap address');
+
+    address newPeer = address(new Peer(_initialSwapContract, _peerOwner));
+    factoryPeers[newPeer] = true;
+
+    emit NewFactoryPeer(newPeer, _initialSwapContract, _peerOwner);
+
+    return newPeer;
```
dmosites: Callers will use the event emitted rather than depending on the return value. Recommend we remove it

examples/peer/contracts/PeerFactory.sol
```diff
@@ -0,0 +1,56 @@
...
+    * @param _peerOwner address that should be the owner of the peer
+    */
+  function deployTrustedPeer(address _initialSwapContract, address _peerOwner) public returns (address) {
+    require(_peerOwner != address(0), 'Provide a peer owner');
+    require(_initialSwapContract != address(0), 'Provide a swap address');
+
+    address newPeer = address(new Peer(_initialSwapContract, _peerOwner));
+    factoryPeers[newPeer] = true;
+
+    emit NewFactoryPeer(newPeer, _initialSwapContract, _peerOwner);
```
dmosites: Would like to have consistency between the function names and events emitted, for example, function `createPeer` emits event `CreatePeer`

examples/peer/contracts/PeerFactory.sol
```diff
@@ -0,0 +1,56 @@
...
+  event NewFactoryPeer(address indexed peer, address swap, address indexed owner);
+
+  mapping(address => bool) internal factoryPeers;
+
+  /**
+    * @notice Deploy a trusted peer contract
+    * @param _initialSwapContract address of the swap contract the peer will deploy with
+    * @param _peerOwner address that should be the owner of the peer
+    */
+  function deployTrustedPeer(address _initialSwapContract, address _peerOwner) public returns (address) {
```
dmosites: Do like the "trusted" language, but it may not be necessary here. The most familiar language for a factory is likely something like `createPeer`.

The addition of `_peerOwner` is good. In the interest of versioning consistency, we should consider removing the `setSwapContract` on the Peer, which simply makes the first argument `_swapContract`. For discussion

protocols/indexer/contracts/Indexer.sol
```diff
@@ -39,19 +40,26 @@ contract Indexer is IIndexer, Ownable {
   // Mapping of address to timestamp of blacklisting
   mapping (address => uint256) public blacklist;
 
+  // The whitelist contract for checking whether a peer is whitelisted
```
dmosites: Whether a locator is whitelisted. Let's rename this to `locatorWhitelist`

dmosites:
Looking good, a few comments. As discussed, we should revisit the `Consumer` example in a later PR and should consider moving this into `protocols/peers`

It may prove useful to interface with a factory in contract code, so recommend we consider breaking out an `interfaces/IPeerFactory.sol`

examples/peer/contracts/PeerFactory.sol
```diff
@@ -0,0 +1,52 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+pragma solidity 0.5.10;
+
+import "@airswap/peer/contracts/Peer.sol";
+import "@airswap/indexer/interfaces/IWhitelist.sol";
+import "@airswap/peer/interfaces/IPeerFactory.sol";
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
```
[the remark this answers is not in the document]
alicevhenshaw: Good spot. Removing.

ejwessel:
My pass is good. Will defer to others to ensure quality.

PR #76, opened 2019-08-14
Simplify indexer

dmosites:
Changes to Indexer to simplify

1. Removing minimum stake amount
2. Removing two-sided market convenience functions
3. Removing lengthOf convenience function
5. Making blacklist a bool rather than timestamp
4. Adding batch blacklisting with arrays of addresses

For (1) this removes the overhead of updating minimum due to volatility. The true minimum emerges from competitive staking. For (2) and (3) they were for convenience and non-essential for basic functionality.


protocols/indexer/contracts/Indexer.sol
```diff
@@ -155,20 +119,21 @@ contract Indexer is IIndexer, Ownable {
...
       "MARKET_IS_BLACKLISTED");
 
     // Ensure the market exists.
     require(markets[_makerToken][_takerToken] != Market(0),
       "MARKET_DOES_NOT_EXIST");
 
-    // Ensure the _amount meets the stakeMinimum.
-    require(_amount >= stakeMinimum,
-      "MINIMUM_NOT_MET");
+    // Only transfer for staking if is amount is set.
```
ejwessel: typo beween `if` and `is`

protocols/indexer/contracts/Indexer.sol
```diff
@@ -30,28 +30,21 @@ contract Indexer is IIndexer, Ownable {
...
   IERC20 public stakeToken;
 
-  // Minimum token amount required for staking
-  uint256 public stakeMinimum;
-
   // Mapping of maker token to taker token to market
   mapping (address => mapping (address => Market)) public markets;
 
   // Mapping of address to timestamp of blacklisting
-  mapping (address => uint256) public blacklist;
```
ejwessel: Curious. Why was this a number before?

dmosites: It was a timestamp that is unnecessary because events are emitted.

dmosites:
@ejwessel updated description with reasoning.

protocols/indexer/contracts/Indexer.sol
ejwessel: as in an address would only be black listed for X duration?

dmosites: Nope it was just the current timestamp.

PR #80, opened 2019-08-15
Remove intent expiries

alicevhenshaw:
### Description

Intents no longer have expiries!


protocols/indexer/contracts/Indexer.sol
```diff
@@ -233,34 +205,32 @@ contract Indexer is IIndexer, Ownable {
...
     address _makerToken,
     address _takerToken,
-    address[] memory _stakers
+    address _staker
   ) internal {
-    for (uint256 i; i < _stakers.length; i++) {
-      if (_stakers[i] != address(0)) {
-        // Get the intent for the sender.
-        Market.Intent memory intent = markets[_makerToken][_takerToken].getIntent(_stakers[i]);
+    if (_staker != address(0)) {
```
poiresel: think it is impossible for _staker to be 0x0 since it's the msg.sender.

alicevhenshaw: yes good spot that remains from cleaning expiries

PR #85, opened 2019-08-16
add unlimited allowance

poiresel:
# Description 

Remove the approve during each transaction and set an unlimited allowance in the constructor.


dmosites:
WETH9.sol decrements approvals on transfers, so this will run out over time. Think we need to stick to existing behavior unfortunately.

dmosites:
Actually WETH9 only decrements if the approval is not `uint(-1)`, so at least for consistency let's use that syntax for the `approve` in the constructor here.

poiresel:
uint(-1) and 2**256 -1 are equivalent. Choosing to keep 2**256 -1 as it's more intuitive that this is the upper limit. A comment has been place in the constructor as well.

PR #86, opened 2019-08-19
Slither Market and Indexer

ejwessel:
`slither Indexer.sol`

handle 
`public-function-that-could-be-declared-as-external`

### DIFF
--- a/protocols/indexer/contracts/Indexer.sol
+++ b/protocols/indexer/contracts/Indexer.sol
@@ -64,7 +64,7 @@ contract Indexer is IIndexer, Ownable {
   function createMarket(
     address _makerToken,
     address _takerToken
-  ) public returns (address) {
+  ) external returns (address) {
 
     // If the Market does not exist, create it.
     if (markets[_makerToken][_takerToken] == Market(0)) {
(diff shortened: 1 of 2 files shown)


protocols/indexer/contracts/Indexer.sol
```diff
@@ -179,10 +179,10 @@ contract Indexer is IIndexer, Ownable {
       "INTENT_DOES_NOT_EXIST");
 
     // Unset the intent on the market.
-    markets[_makerToken][_takerToken].unsetIntent(msg.sender);
+    require(markets[_makerToken][_takerToken].unsetIntent(msg.sender));
```
poiresel: given that market sends back false if there are no intents. I don't think reverting is the correct behavior in this instance.

protocols/indexer/contracts/Indexer.sol
```diff
@@ -179,10 +179,10 @@ contract Indexer is IIndexer, Ownable {
...
 
     // Return the staked tokens.
-    stakeToken.transfer(msg.sender, intent.amount);
+    require(stakeToken.transfer(msg.sender, intent.amount));
```
ejwessel: remove this too.

PR #89, opened 2019-08-20
Update Market and Indexer Logic

ejwessel:
## Description
- Market is coupled to Indexer and intended to be used by the indexer. It is a specialized LL.
- The conditional check in Indexer that check if an intent exists and reverts should be placed into Market.
- Market will continue to sets intents by staker address. Additional logic will now exist if a staker tries to provide another intent, the staking will revert. This keeps with the old design.
- In order to update their intent a stake will need to first remove their old intent and then set a new intent

## Changes
- Update the market so that if a sender were to stake another intent that it will fail. This logic is being moved out of indexer
- Update market and indexer tests to reflect these changes
- Write tests that ensure that the underlying linked list structure has not been compromised.

### DIFF
--- a/protocols/indexer/contracts/Indexer.sol
+++ b/protocols/indexer/contracts/Indexer.sol
@@ -143,11 +143,7 @@ contract Indexer is IIndexer, Ownable {
       // Transfer the _amount for staking.
       require(stakeToken.transferFrom(msg.sender, address(this), _amount),
         "UNABLE_TO_STAKE");
-
     }
-
-    require(!markets[_makerToken][_takerToken].hasIntent(msg.sender),
-      "USER_ALREADY_STAKED");
 
     emit Stake(msg.sender, _makerToken, _takerToken, _amount);
 
--- a/protocols/market/contracts/Market.sol
+++ b/protocols/market/contracts/Market.sol
@@ -107,7 +107,7 @@ contract Market is Ownable {
     bytes32 _locator
   ) external onlyOwner {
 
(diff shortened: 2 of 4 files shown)


ejwessel:
I'm thinking of leaving `hasIntent()` on market as public since it's still useful. But open to making it internal.

protocols/market/test/Market-unit.js
```diff
@@ -358,4 +363,20 @@ contract('Market Unit Tests', async accounts => {
       equal(hasIntent, true, 'hasIntent should have returned true')
     })
   })
+
+  describe('Test multiple setting of intents from same addres', async () => {
```
poiresel: spelling

protocols/indexer/contracts/Indexer.sol
```diff
@@ -143,12 +143,8 @@ contract Indexer is IIndexer, Ownable {
       // Transfer the _amount for staking.
       require(stakeToken.transferFrom(msg.sender, address(this), _amount),
         "UNABLE_TO_STAKE");
```
poiresel: not related to PR, but should be a different message instead of unable_to_stake

protocols/market/contracts/Market.sol
```diff
@@ -173,9 +176,6 @@ contract Market is Ownable {
     // Iterate over the list until the end or limit.
     uint256 i = 0;
     while (i < limit) {
-      if (intent.staker == HEAD) {
-        break;
-      }
```
ejwessel: This code is unreachable.

PR #96, opened 2019-08-22
Migrating Peer from swapSimple to swap

dmosites:
## Description
- rename peerToken and consumerToken -> takerToken and makerToken 
- With the migration to swap, Peer needs to use swap instead.
- remove provideUnsignedOrder as now the function signatures are the same, the difference is whether Signature is populated or not 
- add require to ensure that order.maker.wallet is always msg.sender
- order.maker.kind and order.taker.kind are 0x277f8169 (ERC-20)

### DIFF
--- a/protocols/peer/contracts/Peer.sol
+++ b/protocols/peer/contracts/Peer.sol
@@ -36,6 +36,9 @@ contract Peer is IPeer, Ownable {
   // Mapping of takerToken to makerToken for rule lookup
   mapping (address => mapping (address => Rule)) public rules;
 
+  // ERC-20 (fungible token) interface identifier (ERC-165)
+  bytes4 constant internal ERC20_INTERFACE_ID = 0x277f8169;
+
   /**
     * @notice Contract Constructor
     * @param _swapContract address of the swap contract the peer will deploy with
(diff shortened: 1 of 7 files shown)


protocols/peer/contracts/Peer.sol
```diff
@@ -54,246 +54,222 @@ contract Peer is IPeer, Ownable {
...
-        consumerAmount = peerAmount
+        _makerAmount = _takerAmount
             .mul(rule.priceCoef)
             .div(10 ** rule.priceExp);
 
-        // Ensure that the quoted amount is greater than zero.
-        if (consumerAmount > 0) {
-          return consumerAmount;
-        }
+        // Return the quote.
```
poiresel: nice find!

protocols/peer/contracts/Peer.sol
```diff
@@ -54,246 +54,222 @@ contract Peer is IPeer, Ownable {
...
     return 0;
   }
 
   /**
     * @notice Get a Maximum Quote from the Peer
-    * @param peerToken address The token that the Peer will send
-    * @param consumerToken address The token that the Consumer will send
+    * @param _takerToken address The token that the Peer will send
+    * @param _makerToken address The token that the Consumer will send
     * @return uint 256 The amount the Peer would send
```
poiresel: spacing uint256

PR #112, opened 2019-08-29
Making Market more generic

dmosites:
Removing makerToken and takerToken from Market and other changes to make Market less application-specific.

### DIFF
--- a/protocols/market/contracts/Market.sol
+++ b/protocols/market/contracts/Market.sol
@@ -24,10 +24,6 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
   */
 contract Market is Ownable {
 
-  // Token pair of the market
-  address public makerToken;
-  address public takerToken;
-
   // Length of the linked list
   uint256 public length;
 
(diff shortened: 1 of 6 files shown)


protocols/market/contracts/Market.sol
```diff
@@ -60,34 +56,19 @@ contract Market is Ownable {
     */
 
   event SetIntent(
-    address staker,
-    uint256 amount,
-    bytes32 locator,
-    address makerToken,
-    address takerToken
+    address user,
+    uint256 score,
```
alicevhenshaw: not your fault, but some of these should be indexed

protocols/market/test/Market-unit.js
```diff
@@ -33,8 +30,8 @@ contract('Market Unit Tests', async accounts => {
   const LIST_HEAD = '0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF'
   const LIST_PREV = '0x00'
   const LIST_NEXT = '0x01'
-  const STAKER = 'staker'
-  const AMOUNT = 'amount'
+  const STAKER = 'user'
```
alicevhenshaw: rename the variables to USER and SCORE too

PR #120, opened 2019-08-29
Splitting out PeerFactory into its own package

dmosites:
## Description

Move PeerFactory into its own protocols/ package

## Changes

Move PeerFactory, IPeerFactory, PeerFactory-unit.js  and README into 
~/protocols/peer-factory/ directory
peer-factory is now 0.1.0 package version
peer is now 0.3.0 package version


protocols/peer-factory/migrations/2_deploy_contracts.js
```diff
@@ -0,0 +1,16 @@
...
+const PeerFactory = artifacts.require('PeerFactory')
+const Swap = artifacts.require('@airswap/swap/contracts/Swap.sol')
+const Types = artifacts.require('@airswap/types/contracts/Types.sol')
+
+const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000'
+
+module.exports = deployer => {
+  deployer.deploy(Types)
+  deployer.link(Types, Swap)
+  deployer
```
poiresel: PeerFactory and its tests do not depend on other contracts

PR #125, opened 2019-08-29
rename market to index and intent to signal

poiresel:
# Description

Rename market -> index but still allow Indexer to work with Intents
Rename intent struct -> signal struct
Updated yarn.lock as well


yarn.lock
```diff
@@ -3421,7 +3421,7 @@ eslint-config-airbnb-base@^13.2.0:
   dependencies:
     confusing-browser-globals "^1.0.5"
     object.assign "^4.1.0"
-    object.entries "^1.1.0"
+    object.signals "^1.1.0"
```
alicevhenshaw: should not be renaming things in this file

PR #127, opened 2019-08-30
Terminology update for Index and Signal

dmosites:
## Description

Remove use of intent and instead use locator
From the viewpoint of the indexer, it should work with intents
Only the internal contract index works with locators
Locators are the name of the struct but the actual locator address/bytes32 is stored in the data field
Make sure REVERT_REASON use SIGNAL and not ENTRY

#Test 
Tests all pass with 100% coverage

### DIFF
--- a/protocols/index/contracts/Index.sol
+++ b/protocols/index/contracts/Index.sol
@@ -109,6 +109,7 @@ contract Index is Ownable {
   /**
     * @notice Unset an Locator to Trade
     * @param _user address
+    * @return bool return true on success
     */
   function unsetLocator(
     address _user
@@ -136,6 +137,7 @@ contract Index is Ownable {
   /**
     * @notice Get the Locator for a user
     * @param _user address
+    * @return Locator
     */
   function getLocator(
     address _user
@@ -181,6 +183,7 @@ contract Index is Ownable {
   /**
     * @notice Determine Whether a user is in the Linked List
     * @param _user address
+    * @return bool return true when user exists
     */
   function hasLocator(
     address _user
@@ -196,6 +199,7 @@ contract Index is Ownable {
   /**
     * @notice Returns the first Locator smaller than _score
     * @param _score uint256
+    * @return Locator
     */
   function findPosition(
     uint256 _score
@@ -218,7 +222,7 @@ contract Index is Ownable {
 
   /**
     * @notice Link Two Locators
-    * @dev helper function for linkedlist
+    * @dev helper function for linked list
     * @param _left Locator
     * @param _right Locator
     */
--- a/protocols/index/contracts/Index.sol
+++ b/protocols/index/contracts/Index.sol
@@ -88,7 +88,7 @@ contract Index is Ownable {
     bytes32 _locator
   ) external onlyOwner {
 
-    require(!hasSignal(_user), "USER_HAS_ENTRY");
+    require(!hasSignal(_user), "SIGNAL_ALREADY_SET");
 
     Signal memory newSignal = Signal(_user, _score, _locator);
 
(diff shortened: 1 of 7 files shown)


protocols/index/contracts/Index.sol
```diff
@@ -70,90 +71,90 @@ contract Index is Ownable {
...
+      return locatorsLinkedList[locatorsLinkedList[_user][PREV].user][NEXT];
     }
-    return Signal(address(0), 0, bytes32(0));
+    return Locator(address(0), 0, bytes32(0));
   }
 
   /**
-    * @notice Get Valid Signals
+    * @notice Get Valid Locators
     * @param _count uint256
```
poiresel: may want an @return since return just the data component of the locator

protocols/index/contracts/Index.sol
```diff
@@ -164,67 +165,67 @@ contract Index is Ownable {
...
+      Locator = locatorsLinkedList[Locator.user][NEXT];
     }
-    return intent;
+    return Locator;
   }
 
   /**
-    * @notice Link Two Signals
+    * @notice Link Two Locators
     *
```
poiresel: @dev mention its an helper function that's internal

protocols/indexer/contracts/Indexer.sol
```diff
@@ -131,9 +131,9 @@ contract Indexer is IIndexer, Ownable {
 
     // Ensure both of the tokens are not blacklisted.
     require(!blacklist[_makerToken] && !blacklist[_takerToken],
-      "INDEX_IS_BLACKLISTED");
+      "TOKEN_IS_BLACKLISTED");
```
poiresel: slightly vague message given that TOKEN could be referring to _makerToken or _takerToken.
could just be PAIR_IS_BLACKLISTED

protocols/indexer/test/Indexer-unit.js
```diff
@@ -352,7 +352,7 @@ contract('Indexer Unit Tests', async accounts => {
         indexer.unsetIntent(tokenOne, tokenTwo, {
           from: aliceAddress,
         }),
-        'ENTRY_DOES_NOT_EXIST'
+        'SIGNAL_DOES_NOT_EXIST'
```
poiresel: should be LOCATOR or INTENT just not signal

poiresel: Need to be LOCATOR_DOES_NOT_EXIST

protocols/indexer/test/Indexer.js
```diff
@@ -253,7 +253,7 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
         indexer.unsetIntent(tokenWETH.address, tokenDAI.address, {
           from: aliceAddress,
         }),
-        'ENTRY_DOES_NOT_EXIST'
+        'SIGNAL_DOES_NOT_EXIST'
```
poiresel: LOCATOR_DOES_NOT_EXIST

protocols/indexer/contracts/Indexer.sol
dmosites: It could be either, in which case, we should consider two separate `require` statements. Thoughts?

poiresel: given someone can lookup the token pair freely once they receive the message, since blacklist is public, i don't think having two `require` statements is to helpful.

dmosites: Okay let's go with it.

protocols/index/contracts/Index.sol
dmosites: Makes sense.

PR #141, opened 2019-09-16
remove examples section

poiresel:
## Description

Removing the examples/* section as the contracts will be refactored to better reflect product updates

## Tests

N/A, rest of contracts still pass, only removal no *.sol updates.


alicevhenshaw:
bitmoji

PR #159, opened 2019-10-03
DelegateManager Implementation

ejwessel:
## Description
Reference to : https://github.com/airswap/airswap-protocols/issues/147
The Manager concept is based on the idea to provide users with a way to access all of their created Delegates and simplify the setting of rules and intents.

- [ ] N/A
- [X] New features
- [ ] Bug fixes
- [ ] Typo/small tweaks

## Changes
- Created two new types `Intent` and `Rule` `Types.sol`
- Created `DelegateManager.sol`
- `DelegateManager.sol` Unit Tests

```
 +  DelegateManager (Ownable)
    - [Pub] <Constructor> #
    - [Ext] setRuleAndIntent #
    - [Ext] unsetRuleAndIntent #


 ($) = payable function
 # = non-constant function
```

## Tests
```go
  Contract: DelegateManager Unit Tests
    Test initial values
      ✓ Test factory address
      ✓ Test delegate address
    Test setRuleAndIntent()
      ✓ Test calling setRuleAndIntent with sender token mismatch (177ms)
      ✓ Test calling setRuleAndIntent with sender token mismatch (144ms)
      ✓ Test calling setRuleAndIntent with allowance error (238ms)
      ✓ Test calling setRuleAndIntent with transfer error (308ms)
      ✓ Test calling setRuleAndIntent with approval error (337ms)
      ✓ Test successfully calling setRuleAndIntent (270ms)
    Test unsetRuleAndIntent()
      ✓ Test calling unsetRuleAndIntent() with transfer error (251ms)
      ✓ Test successfully calling unsetRuleAndIntent() (176ms)

  Contract: DelegateManager Integration Tests
    Test initial values
      ✓ Test factory address
      ✓ Test delegate address
    Test setRuleAndIntent()
      ✓ Test successfully calling setRuleAndIntent (275ms)
    Test unsetRuleAndIntent()
      ✓ Test successfully calling unsetRuleAndIntent() (166ms)


  14 passing (5s)

----------------------|----------|----------|----------|----------|----------------|
File                  |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------|----------|----------|----------|----------|----------------|
 contracts/           |      100 |      100 |      100 |      100 |                |
  DelegateManager.sol |      100 |      100 |      100 |      100 |                |
  Imports.sol         |      100 |      100 |      100 |      100 |                |
----------------------|----------|----------|----------|----------|----------------|
All files             |      100 |      100 |      100 |      100 |                |
[...]

Issue #147, opened 2019-09-24
DelegateManager


ejwessel:
Names are based off the renaming proposed here:

[Delegate Renamings](https://github.com/airswap/airswap-protocols/issues/137)

**Note: All names used in this issue will be according to the namings set out in issue 137. However this renaming is not necessary for the concept of the Manager to be understood and discussed. If 137 is not approved, this Manager could be implemented using the old naming conventions**

### Motivation:

- ~~The Manager concept is based on the idea to provide users with a way to access all of their created `Delegates` and~~ simplify the setting of rules and intents.
- Currently a user creates a new `Delegate` using the `createDelegate()` method from the `DelegateFactory`
- Since this `Delegate` is brand new it has no rules and no intents set on `Indexer`.  A user would first need to send a transaction to `setRule()` on the `Delegate` and then subsequently send another transaction to `setIntent()` on the `Indexer`. **Multiple transactions** are necessary to setup the `Delegate`
- Since setting a rule and setting an intent usually go hand in hand it's possible to package both of these in a singular transaction.

### Contract `DelegateManager`:

- *The user interfaces with this contract*
- ~~The manager acts on behalf of a user, keeping track of all the `Delegates` constructed for a particular owner using a `mapping(address ⇒ address[])`~~
- The manager facilitates adding rules to the `Delegate` and `setIntent()` on an `Indexer` in a **single transaction**.
- `DelegateManager` constructor injects `DelgateFactory`

    Functions:

    - ~~`createDelegate()`~~
        - ~~returns a new `Delegate` by calling the `DelegateFactory`. Where the owner of the newly created `Delegate` would be the caller, `msg.sender`.~~
        - ~~Upon creation of a new `Delegate`, the manager saves the address in the owner to list of delegates mapping.~~
    - `constructor()`
        - Upon creation of a new `Delegate`, the manager saves the address
    - `setRuleAndIntent(address Delegate, struct Intent, struct Rule, address Indexer)`
        - sets a `Rule` on a `Delegate` and then subsequently `setIntent()` on an `Indexer` in a **single** transaction.
    - `unsetRuleAndIntent(address Delegate, struct Intent, struct Rule, address Indexer)`
        - unsets a `Rule` on a `Delegate` and then subsequently `unsetIntent()` on an `Indexer` in a **single** transaction.

### Other considerations:

[...]

PR #179, opened 2019-10-22
Delegate Manager Implementation

ejwessel:
## Description
Reference to : https://github.com/airswap/airswap-protocols/issues/147
The Manager concept is based on the idea to provide users with a way to access all of their created Delegates and simplify the setting of rules and intents.

- [ ] N/A
- [X] New features
- [ ] Bug fixes
- [ ] Typo/small tweaks

## Changes
- Created two new types `Intent` and `Rule` `Types.sol`
- Created `DelegateManager.sol`
- `DelegateManager.sol` Unit Tests

```
 +  DelegateManager (Ownable)
    - [Pub] <Constructor> #
    - [Ext] setRuleAndIntent #
    - [Ext] unsetRuleAndIntent #


 ($) = payable function
 # = non-constant function
```

## Tests
```go
  Contract: DelegateManager Unit Tests
    Test initial values
      ✓ Test factory address
      ✓ Test delegate address
    Test setRuleAndIntent()
      ✓ Test calling setRuleAndIntent with allowance error (242ms)
      ✓ Test calling setRuleAndIntent with transfer error (320ms)
      ✓ Test calling setRuleAndIntent with approval error (352ms)
      ✓ Test successfully calling setRuleAndIntent (271ms)
    Test unsetRuleAndIntent()
      ✓ Test calling unsetRuleAndIntent() with transfer error (288ms)
      ✓ Test successfully calling unsetRuleAndIntent() (177ms)

  Contract: DelegateManager Integration Tests
    Test initial values
      ✓ Test factory address
      ✓ Test delegate address
    Test setRuleAndIntent()
      ✓ Test successfully calling setRuleAndIntent (265ms)
    Test unsetRuleAndIntent()
      ✓ Test successfully calling unsetRuleAndIntent() (168ms)


  12 passing (4s)

----------------------|----------|----------|----------|----------|----------------|
File                  |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------|----------|----------|----------|----------|----------------|
 contracts/           |      100 |      100 |      100 |      100 |                |
  DelegateManager.sol |      100 |      100 |      100 |      100 |                |
  Imports.sol         |      100 |      100 |      100 |      100 |                |
----------------------|----------|----------|----------|----------|----------------|
All files             |      100 |      100 |      100 |      100 |                |
----------------------|----------|----------|----------|----------|----------------|
```

Previous PR: https://github.com/airswap/airswap-protocols/pull/159


protocols/delegate-manager/contracts/DelegateManager.sol
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+      * @notice Constructor to create the Delegate Manager
+      * @dev factory is unable to be changed after the DelegateManager is constructed
+      * @param _factory address of the factory contract that will create the delegates
+      * @param _tradeWallet the wallet that the delegate will be trading on behalf of
+      */
+    constructor(IDelegateFactory _factory, address _tradeWallet) public {
+        factory = _factory;
+        delegate = IDelegate(factory.createDelegate(msg.sender, _tradeWallet));
```
ejwessel: Screen Shot 2019-10-22 at 10 16 52 AM

protocols/structs/test/Structs-unit.js
```diff
@@ -0,0 +1 @@
+contract('Types Unit Tests', async () => {})
```
ejwessel: tests in ci fail without this file.

protocols/delegate-manager/contracts/DelegateManager.sol
alicevhenshaw: That would require the Delegate knowing about the indexer though which I'm less keen on. This could almost be combines with the other comment on the PR that alters the factory to deploy the manager too. You could have the following:
`Factory.deployDelegateAndManager`, which performs:
- Deploy manager with msg.sender as the owner
- Deploy delegate with `manager` as the owner

Then in one step all the permissions are setup.

alicevhenshaw:
Please can you link the old PR in the description so its easy to go back and check previous comments?

protocols/delegate-manager/contracts/DelegateManager.sol
ejwessel: > That would require the Delegate knowing about the indexer
Only in context to `setRuleAndIntent` and `unsetRuleAndIntent`. Removing the manager puts the delegate in control of managing itself.

ejwessel:
> Please can you link the old PR in the description so its easy to go back and check previous comments?

added to the bottom of the description

ejwessel:
Few things:
>  struct declarations are only declaring a type not instances of the type.

Perhaps struct.sol is not necessary since structs are types.

Regarding Structs.Rule having `signerToken` and `senderToken`. The tokens can be removed and consolidated in the parameters 
```
  struct Rule {
    uint256 maxSenderAmount;      // The maximum amount of ERC-20 token the delegate would send
    uint256 priceCoef;            // The whole number that will be multiplied by the 10^(-priceExp) - the price coefficient
    uint256 priceExp;             // The exponent of the price to indicate location of the decimal priceCoef * 10^(-priceExp)
  }
```

Regarding Structs.Intent having `signerToken` and `senderToken`. The tokens can be removed and consolidated in the parameters
```
  struct Intent {
    uint256 amount;               // The amount to take
    bytes32 locator;              // Locator data
  }
```
A locator is specific to the user and a locator combines the user with the user's intent.
```
  struct Locator {
    address user;
    Intent userIntent;
  }
```

protocols/types/contracts/Types.sol
```diff
@@ -25,28 +25,34 @@ library Types {
 
   bytes constant internal EIP191_HEADER = "\x19\x01";
 
+  struct Rule {
```
ejwessel: I think this name still collides with the Delegates.

protocols/delegate-manager/migrations/2_deploy_contracts.js
```diff
@@ -0,0 +1,14 @@
...
+
+const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000'
+
+module.exports = deployer => {
+  deployer.deploy(Types)
+  deployer.link(Types, Swap)
+  deployer
+    .deploy(Swap)
+    .then(() => Swap.deployed())
+    .then(() => deployer.deploy(Delegate, Swap.address, EMPTY_ADDRESS, EMPTY_ADDRESS))
```
alicevhenshaw: This doesn't deploy the delegate manager?

protocols/types/contracts/Types.sol
alicevhenshaw: Yeah it's the exact same struct - it needs to be removed from Delegate and imported into Delegate from this library

protocols/delegate-manager/contracts/DelegateManager.sol
```diff
@@ -0,0 +1,121 @@
...
+      require(
+        IERC20(_indexer.stakeToken())
+        .approve(address(_indexer), _amountToStake), "APPROVAL_ERROR"
+      );
+
+      _indexer.setIntent(
+        _signerToken,
+        _senderToken,
+        _amountToStake,
+       bytes32(uint256(address(delegate)) << 96) //NOTE: this will pad 0's to the right
```
alicevhenshaw: I think this needs indenting by a space?

protocols/delegate-manager/contracts/DelegateManager.sol
```diff
@@ -0,0 +1,121 @@
...
+      delegate.setRule(
+        _senderToken,
+        _signerToken,
+        _rule.maxSenderAmount, 
+        _rule.priceCoef,
+        _rule.priceExp
+      );
+
+      require(
+        IERC20(_indexer.stakeToken())
```
alicevhenshaw: You don't need to make this check, as the next require statement will cover it surely?

protocols/delegate-manager/contracts/DelegateManager.sol
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+      * @notice unsets a rule on the delegate and removes an intent on the indexer
+      * @dev delegate needs the manager to be an admin in order to act with it.
+      * @param _senderToken the maker token in the token pair for rules and intents
+      * @param _signerToken the taker token  in the token pair for rules and intents
+      * @param _indexer the indexer to remove the intent from
+      */
+    function unsetRuleAndIntent(
```
alicevhenshaw: Just been thinking... we probably also want a function to change your stake. If someone has a rule for a market, and has staked 50 tokens on that market. If they then want to increase their stake to 100, currently theyd have to call `unsetRuleAndIntent`, which would delete the rule, then they'd have to re-create the rule, and stake again.

Would be easier to have an `changeStake(token, token, newstake)` that calls unset and set for you, without deleting the rule.

protocols/delegate-manager/contracts/DelegateManager.sol
alicevhenshaw: > If we do this it means the DelegateFactory would do more than create delegates.
Yes. But it can just be a slightly more generic factory - for creating Delegates and Managers for Delegates. If we are assuming that people who get Delegates might want a Manager for it too, it makes complete sense to have a function where that all gets set up for them - instead of them having to deploy them separately and then change permissions in them to let them communicate. One function could do all of that.

protocols/delegate-manager/contracts/DelegateManager.sol
ejwessel: There is no explicit reason for the failure when it happens. It just returns
```
Error: Returned error: VM Exception while processing transaction: revert
```
which I didn't feel was descriptive enough from a user standpoint.

protocols/delegate-manager/contracts/DelegateManager.sol
ejwessel: I like this idea. I will look at implementing it today.

ejwessel:
After some discussion we're going to merge the Delegate manager methods into the Delegate for the following reasons:
- no more permission handling 
- DelegateManager is no longer managing multiple delegates
- A Delegate can be returned from the Factory and will have all the methods necessary for interacting with Indexer.

ejwessel:
Now being handled in
https://github.com/airswap/airswap-protocols/pull/189

Will close when the other Pr merges

Issue #180, opened 2019-10-22
Adding starting point for fetching intents


alicevhenshaw:
In the current setup for Indexer and Index, to fetch a set of intents to trade, the caller can specify the number of intents to trade `n`, and they are returned the first `n` intents in the linked list. As the system grows and the linked list gets long, it may be that systems want/need to paginate through the intents in groups. For this reason it would be helpful to allow the starting point to be specified when fetching intents, instead of implicitly starting at the head of the list every time.

Changes:
The following 2 functions each need a second argument `address startPoint` that specifies where the start of the traversal should be.
`Indexer.fetchIntents`
`Index.getLocators`

ejwessel:
```  function getIntents(
    address _signerToken,
    address _senderToken,
    uint256 _count
  )
```
to
```  function getIntents(
    address _signerToken,
    address _senderToken,
    address start
    uint256 _count
  )
```

PR #183, opened 2019-10-22
Intents start point

alicevhenshaw:
## Description
https://github.com/airswap/airswap-protocols/issues/180

Quick description:

- [X] New features

## Tests
All passing


protocols/indexer/contracts/Indexer.sol
```diff
@@ -193,27 +193,29 @@ contract Indexer is IIndexer, Ownable {
...
 
       // Ensure the index exists.
       if (indexes[_signerToken][_senderToken] != Index(0)) {
 
         // Return an array of locators for the index.
-        return indexes[_signerToken][_senderToken].fetchLocators(_count);
+        return indexes[_signerToken][_senderToken].fetchLocators(_startAddress, _count);
 
       }
+
```
ejwessel: Can we clean this method up a bit?

My biggest complaint is the nesting of conditionals. 
We can pull both conditions out and check each one individually before reaching the end where it will return a valid case.

helpers/delegate-frontend/contracts/DelegateFrontend.sol
```diff
@@ -63,12 +63,18 @@ contract DelegateFrontend {
     bytes32[] memory locators = indexer.getIntents(
       _signerToken,
       _senderToken,
+      address(0),
```
ejwessel: can we put a comment here that denotes that we're starting at the head?

protocols/index/test/Index-test.js
```diff
@@ -1,128 +0,0 @@
...
-      assert(locators[0] == aliceLocator, 'Alice should be first')
-      assert(locators[1] == carolLocator, 'Carol should be second')
-      assert(locators[2] == bobLocator, 'Bob should be third')
-      assert(locators[3] == eveLocator, 'Eve should be fourth')
-      assert(locators[4] == davidLocator, 'David should be fifth')
-      assert(locators[5] == zaraLocator, 'Zara should be last')
-      assert(BN(await index.length()).eq(6), 'Index length is incorrect')
-    })
-  })
-})
```
ejwessel: https://github.com/airswap/airswap-protocols/commits/0117c18b2796f78d60ae5c90d020e044bb541aa8/protocols/index/test/Index-test.js

weird history.

protocols/index/test/Index-unit.js
```diff
@@ -17,13 +17,20 @@ contract('Index Unit Tests', async accounts => {
   let bobAddress = accounts[2]
   let carolAddress = accounts[3]
   let davidAddress = accounts[4]
+  let emilyAddress = accounts[5]
+  let fredAddress = accounts[6]
+
+  const HEAD = '0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF'
```
ejwessel: add to `require('@airswap/order-utils').constants` ?

protocols/index/test/Index-unit.js
```diff
@@ -49,8 +56,9 @@ contract('Index Unit Tests', async accounts => {
       let listLength = await index.length()
       equal(listLength, 0, 'Link list length should be 0')
 
-      let locators = await index.fetchLocators(10)
-      equal(locators.length, 0, 'list should have 0 locators')
+      let locators = await index.fetchLocators(EMPTY_ADDRESS, 10)
+      equal(locators.length, 10, 'list should have 10 slots')
```
ejwessel: 10 slots but they're empty?
I guess this is a limitation with being unable to return dynamic arrays. This is odd from a user standpoint. I now have to check the bounds of my returned data because the length can be incorrect. I guess this was how Index was working before. Apologies for not catching this earlier when we were writing Index.

ejwessel: To be clear this is a non issue. This is just a comment.

ejwessel: To be clear this is just a comment.

protocols/index/test/Index-test.js
alicevhenshaw: lol where did this come from

bitmoji

protocols/index/test/Index-unit.js
alicevhenshaw: This was not how the index was working before. Previously you could check which was smaller of `_count` and `length` and create an array of that size before filling it up. This is no longer possible as you aren't starting from the beginning of the list and so you cannot guarantee how long it will be beforehand.

I spoke to Don about it to see if it was an issue because the alternative is creating a new array at the end of the function and copying across the locators one by one, and we concluded this wasnt worth it. Realistically it'll just be used by front ends so they can check

protocols/indexer/contracts/Indexer.sol
alicevhenshaw: I'll just combine into one if statement

alicevhenshaw: https://github.com/airswap/airswap-protocols/pull/183/commits/e6fb1a996d227315bd190ae365225d1b934bce85

protocols/index/test/Index-unit.js
ejwessel: > I guess this was how Index was working before. Apologies for not catching this earlier when we were writing Index.

I don't know why I said this. I swore I saw something earlier - tired eyes.

protocols/indexer/contracts/Indexer.sol
```diff
@@ -193,28 +193,24 @@ contract Indexer is IIndexer, Ownable {
...
-      // Ensure the index exists.
-      if (indexes[_signerToken][_senderToken] != Index(0)) {
-
-        // Return an array of locators for the index.
-        return indexes[_signerToken][_senderToken].fetchLocators(_count);
-
-      }
+    if (!blacklist[_signerToken] && !blacklist[_senderToken] && indexes[_signerToken][_senderToken] != Index(0)) {
+      // Return an array of locators for the index.
+      return indexes[_signerToken][_senderToken].fetchLocators(_startAddress, _count);
```
ejwessel: Sorry, I know this is me being picky, but I was thinking more of a fail first style. 

i.e. The first condition should return `bytes32[](0)` and where the valid case is at the end. This way if everything is valid the code falls through to pass.

protocols/indexer/contracts/Indexer.sol
```diff
@@ -193,28 +193,24 @@ contract Indexer is IIndexer, Ownable {
...
   function getIntents(
     address _signerToken,
     address _senderToken,
+    address _startAddress,
     uint256 _count
   ) external view returns (
     bytes32[] memory locators
   ) {
-
     // Ensure neither token is blacklisted.
```
ejwessel: add the deleted comment

```
// Ensure the index exists.
```

protocols/indexer/contracts/Indexer.sol
alicevhenshaw: Will change

alicevhenshaw: image

PR #193, opened 2019-10-24
Updated security report and slither output for Wrapper

dmosites:


helpers/wrapper/reports/SECURITY.md
```diff
@@ -1,150 +1,80 @@
-# Security Report: Wrapper
+# Wrapper: Security Report
 
-Smart Contract Security Report by Team Fluidity (team[at]fluidity[dot]io) and Phil Daian (feedback[at]stableset[dot]com)
-Hash of master used for report: [ef5cff0613532d27ecedb332e222ae0a75079841](https://github.com/airswap/airswap-protocols/commit/ef5cff0613532d27ecedb332e222ae0a75079841)
+Security report by Team Fluidity (team[at]fluidity[dot]io). Smart contracts are a nascent space, and no security audit procedure has been perfected. We welcome any suggestions and comments on this report, its contents, our methodology, or potential gaps in coverage.
```
alicevhenshaw: Are we having a bounty? If so could mention that here, just with a link to where they can find more info on a bug bounty?

helpers/wrapper/reports/SECURITY.md
```diff
@@ -1,150 +1,80 @@
...
 
 ## Introduction
 
-The Swap Protocol is a peer-to-peer protocol for trading Ethereum tokens that allows two parties to exchange tokens in an atomic transaction. It is a non-custodial exchange settlement contract. The "Wrapper" contract facilitates sending and receiving ether for WETH (wrapped ether) trades on the Swap contract. Contracts are compiled with v0.5.10.a6ea5b19 (0.5.10 stable release).
+Wrapper is a frontend for a Swap contract to enable sending and receiving ether (ETH) for wrapped ether (WETH) trades. The Swap contract only works with tokens, so WETH is required. Wrapper is written and primarily intended for the AirSwap Instant system. If an order does not specify WETH for either party, execution is passed directly through to the Swap contract. The following contracts are compiled with solidity 0.5.12.
 
 ## Structure
 
-Wrapper is comprised of a single contract.
+Wrapper includes one contract and its dependencies.
```
alicevhenshaw: `Wrapper includes one contract` I prefer how it used to be - _includes_ is a weird word, it feels like the one contract is included within wrapper, but its not, wrapper IS that contract.

helpers/wrapper/reports/SECURITY.md
```diff
@@ -1,150 +1,80 @@
...
+Wrapper [Source Code](https://github.com/airswap/airswap-protocols/tree/master/helpers/wrapper) and [README](../README.md) are available in this repository. Commit used for report: [2a83c1ff2e46e6befa45889aa556fdd31e5c71fb](https://github.com/airswap/airswap-protocols/commit/2a83c1ff2e46e6befa45889aa556fdd31e5c71fb)
 
 ## Introduction
 
-The Swap Protocol is a peer-to-peer protocol for trading Ethereum tokens that allows two parties to exchange tokens in an atomic transaction. It is a non-custodial exchange settlement contract. The "Wrapper" contract facilitates sending and receiving ether for WETH (wrapped ether) trades on the Swap contract. Contracts are compiled with v0.5.10.a6ea5b19 (0.5.10 stable release).
+Wrapper is a frontend for a Swap contract to enable sending and receiving ether (ETH) for wrapped ether (WETH) trades. The Swap contract only works with tokens, so WETH is required. Wrapper is written and primarily intended for the AirSwap Instant system. If an order does not specify WETH for either party, execution is passed directly through to the Swap contract. The following contracts are compiled with solidity 0.5.12.
```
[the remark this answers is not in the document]
alicevhenshaw: I think the intro should have a sentence stating that the ETH part of the trade is only one sided. i.e. if using the wrapper, one side of the party is still going to be sending/receiving WETH, just one party will benefit from the use of the wrapper

helpers/wrapper/reports/SECURITY.md
```diff
@@ -1,150 +1,80 @@
...
-
-#### 11. Wrapper should not have any impact on affiliates specified to receive WETH.
-
-- In this scenario, an affiliate that is specified to receive WETH will receive WETH if Swap.swap succeeds. There is no wrapping or unwrapping for affiliate parties.
-- **This invariant currently holds as-is.**
-
-## Analysis
-
-#### Slither (static analyzer)
+- By inspection and testing, line 70 requires that the message sender (transaction originator) is the same as the order sender wallet.
```
alicevhenshaw: message sender != transaction originator
if using a multisig, the message sender would be the multisig, but the transaction originator would be an EOA

helpers/wrapper/reports/slither.txt
```diff
@@ -0,0 +1,37 @@
...
+	- Full.sol#20 declares pragma solidity0.5.12
+	- Full.sol#21 declares pragma experimentalABIEncoderV2
+	- Full.sol#172 declares pragma solidity0.5.12
+	- Full.sol#295 declares pragma solidity0.5.12
+	- Full.sol#314 declares pragma solidity^0.5.0
+	- Full.sol#409 declares pragma solidity0.5.12
+Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#different-pragma-directives-are-used
+INFO:Detectors:
+Pragma version "0.5.12" allows old versions (Full.sol#20)
+Pragma version "0.5.12" allows old versions (Full.sol#172)
```
alicevhenshaw: I know this is a slither output. Question about the actual output, not really the PR...
Does it? I thought the lack of `^` meant the version was fixed?

helpers/wrapper/reports/SECURITY.md
dmosites: Bounty will be published next week and can be added to these reports when ready

alicevhenshaw:
Happy with these invariants. Having reviewed I think they cover everything. Just a couple of thoughts of things we should potentially mention:
- We didn't write the WETH contract ourselves. However it's publicly verified on etherscan, and is not upgradeable and so we trust it
- The contract does not work for trading ETH for WETH, as this is denoted as WETH for WETH in the order, and therefore the contract tries to both wrap and unwrap. It's not a security problem persay, just if people were to try to use it for this it would behave unexpectedly. However it's not intended for this usecase, and this usecase will not be possible on our UI.

helpers/wrapper/reports/slither.txt
dmosites: It appears to complain if any version other than `0.5.3` is set. We should consider adding the caret to these to keep valid compilers under the `0.6.0`.

dmosites:
> * We didn't write the WETH contract ourselves. However it's publicly verified on etherscan, and is not upgradeable and so we trust it

Wrapper does not use a specific WETH contract, only an IWETH interface we developed

> * The contract does not work for trading ETH for WETH, as this is denoted as WETH for WETH in the order, and therefore the contract tries to both wrap and unwrap. It's not a security problem persay, just if people were to try to use it for this it would behave unexpectedly. However it's not intended for this usecase, and this usecase will not be possible on our UI.

Will add a note about this behavior

PR #215, opened 2019-10-28
added small updates to wrapper security report

poiresel:
## Description

Small updates for the wrapper update
Quick description:

- [X] Documentation changes

## Changes
Updated the wrapper.security.md to better reflect current repo state.

## Tests
No changes

## Test Coverage

No changes

### DIFF
--- a/source/wrapper/reports/SECURITY.md
+++ b/source/wrapper/reports/SECURITY.md
@@ -114,8 +114,6 @@ Wrapper.swap(Types.Order) (Full.sol#456-506) ignores return value by external ca
 Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#unused-return
 ```
 
-The WETH contract that is used is known contract with its behavior well understood. The WETH contract will either return true or revert if there is a failure. As far as the code describes, there is no way that false could be returned or that true would be returned without the specified amount of WETH tokens transferred.
-
 ## Notes
 
 - When deploying a Wrapper contract, we recommend using the [canonical WETH contracts](https://blog.0xproject.com/canonical-weth-migration-8a7ab6caca71).
--- a/source/wrapper/reports/SECURITY.md
+++ b/source/wrapper/reports/SECURITY.md
@@ -27,11 +27,20 @@ contracts/Wrapper.sol
 `†` [@airswap/tokens](https://github.com/airswap/airswap-protocols/tree/master/source/tokens)
 `‡` [Open Zeppelin v2.0 Security Audit](https://drive.google.com/file/d/1gWUV0qz3n52VEUwoT-VlYmscPxxo9xhc/view)
 
+## Constructor Params
+
+```
+`*` [@airswap/swap](https://github.com/airswap/airswap-protocols/tree/master/source/swap)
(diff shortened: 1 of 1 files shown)


source/wrapper/reports/SECURITY.md
```diff
@@ -61,7 +70,7 @@ contracts/Wrapper.sol
 
 ### All expected behavior of the Swap contract holds entirely.
 
-- By testing, the underlying `swap` function is executed and its variants hold as expected.
+- By testing, the underlying `swap` function is executed and its invariants hold as expected.
```
alicevhenshaw: good catch

source/wrapper/reports/SECURITY.md
```diff
@@ -105,6 +114,8 @@ Wrapper.swap(Types.Order) (Full.sol#456-506) ignores return value by external ca
 Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#unused-return
 ```
 
+The WETH contract that is used is known contract with its behavior well understood. The WETH contract will either return true or revert if there is a failure. As far as the code describes, there is no way that false could be returned or that true would be returned without the specified amount of WETH tokens transferred.
```
alicevhenshaw: Back on the original PR adding this report to the repo, Don and I had the following convo:
Alice:
> Happy with these invariants. Having reviewed I think they cover everything. Just a couple of thoughts of things we should potentially mention:
> - We didn't write the WETH contract ourselves. However it's publicly verified on etherscan, and is not upgradeable and so we trust it

Don:
> Wrapper does not use a specific WETH contract, only an IWETH interface we developed. Will add a note to suggest that users use the canonical WETH contracts on rinkeby and mainnet.

From this conversation, I'm therefore not sure this addition is necessarily correct - as we don't enforce a specific WETH contract in wrapper? What are your thoughts?

poiresel: If that's the case then we should put a require on the transfer and transferFrom then in the wrapper contract as that statement implies we're treating WETH like a black box and then could have unexpected behavior.

alicevhenshaw: In my opinion, we hardcode the weth address so this would be an unnecessary addition. I view it as WETH not IWETH. But let's get @dmosites thoughts on this.

poiresel: i would like to use WETH9.sol then if that's the path forward.

dmosites: Mentioned in the report we should consider adding `require` statements. While we do not use a specific WETH contract, though, I've come to think that it's worth mentioning the WETH contract that we _do_ use for testing and later for deployments, as that's a part of our security posture. For the above addition, I might qualify it as, "The WETH contract used for testing and deployments by the AirSwap team is a known [...]

poiresel: ok, i may have missed that section of the report, I incorrectly thought "If possible, the team should consider ensuring expected behavior with `require` statements." came from Slither.

poiresel: just to close this, i'll remove my accessory comment as it sounds like this has been thought about by the team and just a note in the Notes section is sufficient.

dmosites: "The WETH contract that is used is **a** known contract [...["

PR #222, opened 2019-10-28
added in usdt and cryptokitty support

poiresel:
## Description

Quick description:

- [X] N/A
- [ ] New features
- [ ] Bug fixes
- [ ] Typo/small tweaks

## Changes

Implementation details. These are mainly to help your fellow developers.

### DIFF
--- a/source/swap/contracts/Swap.sol
+++ b/source/swap/contracts/Swap.sol
@@ -18,13 +18,15 @@ pragma solidity 0.5.12;
 pragma experimental ABIEncoderV2;
 
 import "@airswap/swap/contracts/interfaces/ISwap.sol";
-import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "@airswap/tokens/contracts/interfaces/IERC20.sol";
 import "openzeppelin-solidity/contracts/token/ERC721/IERC721.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
 /**
   * @title Swap: The Atomic Swap used by the Swap Protocol
   */
 contract Swap is ISwap {
+  using SafeMath for uint256;
 
   // Domain and version for use in signatures (EIP-712)
   bytes constant internal DOMAIN_NAME = "SWAP";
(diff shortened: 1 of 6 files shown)


source/swap/contracts/Swap.sol
```diff
@@ -335,10 +337,12 @@ contract Swap is ISwap {
   ) internal {
     if (kind == ERC721_INTERFACE_ID) {
       // Attempt to transfer an ERC-721 token.
-      IERC721(token).safeTransferFrom(from, to, param);
+      IERC721(token).transferFrom(from, to, param);
```
alicevhenshaw: Do we need to document this somewhere so that the frontend check gets picked up by the frontend devs?

poiresel: yes, jotting these down in the issue + notion doc and will share

source/swap/contracts/Swap.sol
```diff
@@ -335,10 +339,13 @@ contract Swap is ISwap {
...
     } else {
-      // Attempt to transfer an ERC-20 token.
-      require(IERC20(token).transferFrom(from, to, param));
+      require(from != to, "TO_CANNOT_EQUAL_FROM");
```
alicevhenshaw: Should this check be taken out the front of the if statement? Does it make sense to have IERC721 transferring tokens with from and to the same?

source/swap/contracts/Swap.sol
```diff
@@ -335,10 +339,13 @@ contract Swap is ISwap {
...
+      // Attempt to transfer an ERC-20 token and checks to begin and end balacnes.
```
alicevhenshaw: Typos. Maybe
`      // Attempt to transfer an ERC-20 token with checks to balances before and after.`

source/swap/contracts/Swap.sol
poiresel: I can move it up, just initially was trying to keep it contained.

source/swap/contracts/Swap.sol
poiresel: thanks!

source/swap/test/Swap-unit.js
```diff
@@ -114,6 +116,67 @@ contract('Swap Unit Tests', async accounts => {
...
+      const token_balance = fungibleTokenTemplate.contract.methods
+        .balanceOf(EMPTY_ADDRESS)
+        .encodeABI()
+
+      const token_transfer = fungibleTokenTemplate.contract.methods
+        .transferFrom(EMPTY_ADDRESS, EMPTY_ADDRESS, 0)
+        .encodeABI()
+
+      // The token transfer should return true
+      await tokenMock.givenMethodReturnBool(token_transfer, true)
```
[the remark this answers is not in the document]
poiresel: sure, will do.

Issue #225, opened 2019-10-29
Indexer.updateStakeAmount function


alicevhenshaw:
Currently if a user wants to update their amount staked on a market, they have to call `unsetIntent` and then `setIntent`, which requires them signing and sending 2 transactions. Considering the indexer will revolve around only being able to return the first n staked locators, it should be expected that people might want to update their stake to ensure they remain competitive. For this reason I propose a function `updateStakeAmount(signerToken, senderToken, newAmount` that updates the user's staked amount. This function need not be complex - it could merely call `unsetIntent` and `setIntent` internally, but it stops the user from having to sign 2 transactions.

dmosites:
Also for consideration: `setIntent` will `unsetIntent` if one exists before setting the new intent.

alicevhenshaw:
Yes @poiresel, I would say that it could be an increase or a decrease, and tokens are transferred either way.

PR #232, opened 2019-10-29
Indexer update stake

alicevhenshaw:
## Description

Quick description:

- [X] New features

## Changes

- Updated `setIntent` to allow stakes and locators to be updated without unsetting intent first. Discussed in issue https://github.com/airswap/airswap-protocols/issues/225.
- Added tests for this test case, as well as some more tests for the locatorWhitelist, as this was seemingly lacking in the integration tests previously.

## Tests
  90 passing (15s)

## Test Coverage
100%

### DIFF
--- a/source/index/contracts/Index.sol
+++ b/source/index/contracts/Index.sol
@@ -168,7 +168,7 @@ contract Index is Ownable {
     result = new bytes32[](count);
 
     // Iterate over the list until the end or count.
-    uint8 i = 0;
+    uint256 i = 0;
     while (i < count && identifier != HEAD) {
       result[i] = entries[identifier].locator;
       i = i + 1;
--- a/source/index/contracts/Index.sol
+++ b/source/index/contracts/Index.sol
@@ -139,7 +139,7 @@ contract Index is Ownable {
     */
   function getLocator(
     address identifier
-  ) external view returns (uint256) {
+  ) external view returns (bytes32) {
     return entries[identifier].locator;
   }
 
--- a/source/index/contracts/Index.sol
+++ b/source/index/contracts/Index.sol
@@ -132,6 +132,17 @@ contract Index is Ownable {
     return entries[identifier].score;
   }
 
+    /**
+    * @notice Get a Locator
+    * @param identifier address On-chain address identifying the owner of a locator
+    * @return uint256 Locator information
+    */
+  function getLocator(
+    address identifier
+  ) external view returns (uint256) {
+    return entries[identifier].locator;
+  }
+
   /**
     * @notice Get a Range of Locators
     * @dev start value of 0x0 starts at the head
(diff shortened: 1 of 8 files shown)


source/index/contracts/Index.sol
```diff
@@ -157,7 +168,7 @@ contract Index is Ownable {
     result = new bytes32[](count);
 
     // Iterate over the list until the end or count.
-    uint8 i = 0;
+    uint256 i = 0;
```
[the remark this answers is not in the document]
poiresel: think you can just put uint256 i; since default is 0

source/indexer/test/Indexer.js
```diff
@@ -242,9 +358,57 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
...
+      emitted(
+        await indexer.unsetIntent(tokenWETH.address, tokenDAI.address, {
+          from: bobAddress,
+        }),
+        'Unstake'
+      )
+    })
+
+    it('Remove locator whitelist', async () => {
+      await indexer.setLocatorWhitelist(EMPTY_ADDRESS, {
```
poiresel: Does this need to be checked that it was removed and set to 0x0?

source/indexer/test/Indexer.js
```diff
@@ -231,9 +263,93 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
...
+          }
+        ),
+        'Stake'
+      )
+
+      // Bob has 400 fewer AST, now the indexer owns them
+      ok(await balances(bobAddress, [[tokenAST, 600]]))
+      ok(await balances(indexerAddress, [[tokenAST, 400]]))
+
+      let staked = await indexer.getStakedAmount.call(
```
poiresel: have an example where stakeAmount does not change.

source/indexer/contracts/Indexer.sol
```diff
@@ -277,6 +272,50 @@ contract Indexer is IIndexer, Ownable {
...
+    address user,
+    address signerToken,
+    address senderToken,
+    uint256 newAmount,
+    bytes32 newLocator,
+    uint256 oldAmount
+  ) internal {
+    // If the new stake is bigger, collect the difference.
+    if (oldAmount < newAmount) {
+      require(stakingToken.transferFrom(user, address(this), newAmount - oldAmount),
```
poiresel: safeMath should be used.

source/indexer/contracts/Indexer.sol
```diff
@@ -277,6 +272,50 @@ contract Indexer is IIndexer, Ownable {
...
+        "UNABLE_TO_STAKE");
+    }
+
+    // If the old stake is bigger, return the excess.
+    if (newAmount < oldAmount) {
```
poiresel: note that safemath not needed to be used since overflow is done by the less than check

Issue #224, opened 2019-10-29
Return the score and identifier from Index.getLocators or add comparable call


dmosites:
When using the Indexer, the score and identifier are also important information to have when fetching locators.

- `score`: The user needs to know how much they must to stake to rank on the Index.
- `identifier`: The user needs to have the identifier of at least the final entry for pagination.

They should be returned in the same call to represent the same state.

A route may be to return e.g. `(bytes32[], uint256[], address[])` from the `getLocators` call for locators, scores, and identifiers respectively. Alternatively, if only the last entry is useful for pagination, `(bytes32[], uint256[], address)`.

poiresel:
@dmosites can you give a user flow for this, I'm not 100% understanding it?

PR #242, opened 2019-10-30
Get locators update

alicevhenshaw:
## Description

PR for https://github.com/airswap/airswap-protocols/issues/224.

## Changes

- `getLocators` in `Indexer` and `Index` to return scores, locators and the next user's address. Previously it only returned locators.
- Additionally, the upper bound of array size in `Index.getLocators` has been tightened

## Tests
All passing


source/index/contracts/Index.sol
```diff
@@ -148,13 +148,18 @@ contract Index is Ownable {
...
     */
   function getLocators(
     address start,
     uint256 count
-  ) external view returns (bytes32[] memory result) {
-
+  ) external view returns (
+    bytes32[] memory locators,
+    uint256[] memory scores,
+    address nextIdentifier
```
poiresel: should nextIdentifier be the name of the return term in line 189?

alicevhenshaw: I thought `nextIdentifier` was more descriptive of what was being returned to the called than `identifier`.

I could call it `nextIdentifier` throughout the function though

poiresel: yes, and i think that's accurate in terms how that variable is called. the initialized value is `entries[HEAD].next;`

alicevhenshaw:
@poiresel Index.getScore also exists so I think that would be superfluous. Though we could combine them to one. Thoughts?

poiresel:
> @poiresel Index.getScore also exists so I think that would be superfluous. Though we could combine them to one. Thoughts?

That works since they are both quick retrievals.

source/index/contracts/Index.sol
```diff
@@ -11,7 +11,7 @@
   distributed under the License is distributed on an "AS IS" BASIS,
   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   See the License for the specific language governing permissions and
-  limitations under the License.
+  sizeations under the License.
```
poiresel: bad copy and replace?

alicevhenshaw: hahahah yes. Will revert it.

source/index/contracts/Index.sol
```diff
@@ -147,33 +147,47 @@ contract Index is Ownable {
     * @notice Get a Range of Locators
     * @dev start value of 0x0 starts at the head
     * @param start address Identifier to start with
-    * @param count uint256 Number of locators to return
-    * @return bytes32[] result List of locators
+    * @param limit uint256 Number of locators to return
+    * @return bytes32[] List of locators
+    * @return uint256[] List of scores corresponding to locators
+    * @return address The next identifier to provide for pagination
```
poiresel: next cursor? not sure if that's the right usage of the term but replace with something that's not identifier.

utils/order-utils/src/constants.js
```diff
@@ -20,6 +20,9 @@ module.exports = {
   SECONDS_IN_DAY: 86400,
   EMPTY_ADDRESS: '0x0000000000000000000000000000000000000000',
   HEAD: '0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF',
```
poiresel: why are some of the f's lower case and some upper case?

source/indexer/test/Indexer.js
```diff
@@ -495,7 +512,7 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
     })
 
     it('Bob ensures there are no more intents the Indexer', async () => {
-      const intents = await indexer.getLocators.call(
+      result = await indexer.getLocators.call(
         tokenWETH.address,
         tokenDAI.address,
         EMPTY_ADDRESS,
```
poiresel: can you make 10 a constant like DEFAULT_QUERY_SIZE or something?

utils/order-utils/src/constants.js
alicevhenshaw: Its a check summed Ethereum address

source/index/contracts/Index.sol
```diff
@@ -146,34 +146,49 @@ contract Index is Ownable {
   /**
     * @notice Get a Range of Locators
     * @dev start value of 0x0 starts at the head
-    * @param start address Identifier to start with
-    * @param count uint256 Number of locators to return
-    * @return bytes32[] result List of locators
+    * @param cursor address Cursor to start with
+    * @param limit uint256 Number of locators to return
```
dmosites: *Maximum number of locators to return

source/index/contracts/Index.sol
```diff
@@ -146,34 +146,49 @@ contract Index is Ownable {
...
+    bytes32[] memory locators,
+    uint256[] memory scores,
+    address nextCursor
+  ) {
+    address identifier;
+
+    // If a valid cursor is provided, start there.
+    if (cursor != address(0) && cursor != HEAD) {
+      // Check that the provided cursor exists.
+      require(_hasEntry(cursor), "START_ENTRY_NOT_FOUND");
```
dmosites: `CURSOR_NOT_FOUND`

source/indexer/test/Indexer-unit.js
```diff
@@ -586,17 +594,19 @@ contract('Indexer Unit Tests', async accounts => {
   })
 
   describe('Test getLocators', async () => {
-    it('should return an empty array if the index doesnt exist', async () => {
-      let intents = await indexer.getLocators.call(
+    it('should return blank results if the index doesnt exist', async () => {
+      result = await indexer.getLocators.call(
```
ejwessel: Is the global supposed to be the `let result` here?

source/indexer/test/Indexer.js
```diff
@@ -30,6 +36,8 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
   let tokenDAI
   let tokenWETH
 
+  let result
```
ejwessel: global question again.

ejwessel:
Other than my comments on result. This looks great.

source/index/test/Index-unit.js
```diff
@@ -23,6 +29,8 @@ contract('Index Unit Tests', async accounts => {
   let snapshotId
   let index
 
+  let result
```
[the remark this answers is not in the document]
alicevhenshaw: just global because it got used so many times in so many tests

PR #248, opened 2019-10-31
Adding latest deploys and bumping versions

dmosites:


source/delegate-factory/README.md
```diff
@@ -18,11 +18,11 @@
 - Blog → https://blog.airswap.io/
 - Support → https://support.airswap.io/
 
-## Deployments
+## Deploys
```
alicevhenshaw: Why this change? I way prefer `deployments`. Though I see this isn't important right now.

Issue #251, opened 2019-10-31
Ensure that there are no extraneous dependencies, or dependencies that belong in devDependencies


dmosites:
Currently, for example, `indexer` lists `solidity-coverage` as a dependency, when it should be a devDependency.

poiresel:
delegate  needs to be updated as well

Below shows change
  "devDependencies": {
    "@airswap/order-utils": "0.3.3",
    "@airswap/test-utils": "0.1.0",
    "solidity-coverage": "^0.6.3"
  },
  "dependencies": {
    "@airswap/swap": "2.2.0",
    "openzeppelin-solidity": "^v2.2"
  }

poiresel:
Update for delegate-factory 

 "devDependencies": {
    "@airswap/order-utils": "0.3.0",
    "@airswap/test-utils": "0.1.0",
    "solidity-coverage": "^0.6.3"
  },
  "dependencies": {
    "@airswap/delegate": "0.4.0"
  }

poiresel:
Update for index

  "devDependencies": {
    "@airswap/order-utils": "0.3.3",
    "@airswap/test-utils": "0.1.0",
    "@airswap/tokens": "0.1.0",
    "solidity-coverage": "^0.6.3",
    "truffle": "^5.0.24",
    "bignumber.js": "^9.0.0"
  },
  "dependencies": {
    "openzeppelin-solidity": "^v2.2"
  }

poiresel:
#256 Closing as this has been addressed.

PR #256, opened 2019-10-31
Moving some dependencies to devDependencies

dmosites:
## Description

Implements #251. Moving some dependencies to devDependencies

- [X] Typo/small tweaks

## Changes

Changes to package.json in affected packages.

### DIFF
--- a/source/indexer/test/Indexer.js
+++ b/source/indexer/test/Indexer.js
@@ -203,7 +203,7 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
             from: aliceAddress,
           }
         ),
-        'SafeMath: subtraction overflow'
+        'ERC20: transfer amount exceeds balance'
       )
     })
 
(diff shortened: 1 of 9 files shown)


source/indexer/test/Indexer.js
```diff
@@ -203,7 +203,7 @@ contract('Indexer', async ([ownerAddress, aliceAddress, bobAddress]) => {
             from: aliceAddress,
           }
         ),
-        'SafeMath: subtraction overflow'
+        'ERC20: transfer amount exceeds balance'
```
poiresel: was this failing?

yarn.lock
```diff
@@ -18,16 +18,15 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/generator@^7.5.5":
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.5.5.tgz#873a7f936a3c89491b43536d12245b626664e3cf"
-  integrity sha512-ETI/4vyTSxTzGnU2c49XHv2zhExkv9JHLTwDAFz85kmcwuShvYG2H08FwgIguQf4JC75CBnXAUM5PqeF4fj0nQ==
+"@babel/generator@^7.6.3":
```
alicevhenshaw: For what reason has this change led to 2000+ lines updated on yarn? Aren't all the dependencies the same?

alicevhenshaw:
We should consider locking down our versions of everything not just open zeppelin. The others are less risky as they dont go into production.

PR #263, opened 2019-11-01
Delegate event

alicevhenshaw:
## Description

Added an event to Delegate.provideOrder, outputted on successful swap.

Quick description:

- [X] N/A

## Tests
All passing

## Test Coverage
All passing

### DIFF
--- a/source/delegate/contracts/interfaces/IDelegate.sol
+++ b/source/delegate/contracts/interfaces/IDelegate.sol
@@ -42,6 +42,15 @@ interface IDelegate {
     address indexed signerToken
   );
 
+  event Trade(
+    address indexed ruleOwner,
+    address indexed senderToken,
+    address indexed signerToken,
+    uint256 senderAmount,
+    uint256 priceCoef,
+    uint256 priceExp
+  );
+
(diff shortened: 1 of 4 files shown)


source/delegate/contracts/interfaces/IDelegate.sol
```diff
@@ -42,6 +42,15 @@ interface IDelegate {
     address indexed signerToken
   );
 
+  event ProvideOrder(
+    address indexed ruleOwner,
```
dmosites: `delegateOwner`?

alicevhenshaw: its called `delegateOwner` in the other events currently. (setrule and unsetrule) but could rename it in those too

dmosites: I see. Maybe similar to `staker` elsewhere this could just be `owner` in both cases.

poiresel: Did you think it was worth having both owner + tradeWallet but indexing by owner?

Issue #228, opened 2019-10-29
Add "version" to Order object and get*Order calls to indicate the Swap contract intended for settlement


dmosites:
Given that there may be many Swap contracts, whether for different systems, or through incremental versions, two peers will need to agree on which contract they intend to settle trades.

The value of `version` would be a Swap contract address.

alicevhenshaw:
This seems fair

dmosites:
Yes, sorry. Meant to say that the value of `version` would be a Swap contract address.

alicevhenshaw:
Ah I see. I thought you meant the `DOMAIN_VERSION` constant that already exists.

poiresel:
If that's the case, then maybe modify the Types.hashDomain to use the order.version if that's the new verifying contract?

dmosites:
> If that's the case, then maybe modify the Types.hashDomain to use the order.version if that's the new verifying contract? Also recommend calling it versioningAddress / versioningContract or something else appending to indicate this is a contractAddress and not just a a version 1 or 2

The EIP-712 `domainSeparator` is calculated by `hashDomain` at deploy time for each Swap contract, not on a per-order basis. As for naming, agree it should be clear, though it could be clear simply to see that the values and types are addresses.

poiresel:
dmosites 3:03 PM
signatureValidator is flat
signature.validator is nested
it would be a part of Signature struct

poiresel:
Close as #265 has been merged

PR #265, opened 2019-11-04
add validator to Signature struct

poiresel:
## Description

Quick description:
- [x] New features

#228 

## Changes

Core change: addition of validator to Types.sol

```
  struct Signature {
    address signatory;            // Address of the wallet used to sign
    address validator;            // Address of the intended swap contract
    bytes1 version;               // EIP-191 signature version
    uint8 v;                      // `v` value of an ECDSA signature
    bytes32 r;                    // `r` value of an ECDSA signature
    bytes32 s;                    // `s` value of an ECDSA signature
  }
```

Smaller changes

Upgrade `package.json` versions that weren't initially upgraded
Remove `NULL_ADDRESS` constant usage as it does not exist

## Tests

##Test Coverage

### DIFF
--- a/source/indexer/package.json
+++ b/source/indexer/package.json
@@ -19,7 +19,7 @@
   "devDependencies": {
     "@airswap/index": "0.3.0",
     "@airswap/test-utils": "0.1.0",
-    "@airswap/types": "0.3.0",
+    "@airswap/types": "0.3.1",
     "bignumber.js": "^9.0.0",
     "solidity-coverage": "^0.6.3",
     "truffle": "^5.0.24"
--- a/source/indexer/package.json
+++ b/source/indexer/package.json
@@ -17,7 +17,7 @@
     "test": "truffle test"
   },
   "devDependencies": {
-    "@airswap/index": "0.3.0",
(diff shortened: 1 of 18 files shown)


source/indexer/package.json
```diff
@@ -19,7 +19,7 @@
   "devDependencies": {
     "@airswap/index": "0.3.0",
```
alicevhenshaw: Index is on 0.3.1 too https://www.npmjs.com/package/@airswap/index

source/swap/contracts/Swap.sol
```diff
@@ -74,6 +74,10 @@ contract Swap is ISwap {
     Types.Order calldata order
   ) external {
 
+    // Ensure signatureVerifier defined in order is this contract
+    require(order.signatureVerifier == address(this),
+      "INVALID_SWAP_ADDRESS");
```
alicevhenshaw: I guess if we're naming the variable `signatureVerifier`, then this revert message should change too

poiresel: removing the check per dons's comment: "it doesn't need to be checked because the contract address is in the domainSeparator"

PR #269, opened 2019-11-05
fixing tokens version

alicevhenshaw:
## Description

Updating version of airswap tokens used to the correct one.

### DIFF
--- a/source/index/package.json
+++ b/source/index/package.json
@@ -19,7 +19,7 @@
   "devDependencies": {
     "@airswap/order-utils": "0.3.3",
     "@airswap/test-utils": "0.1.1",
-    "@airswap/tokens": "0.1.1",
+    "@airswap/tokens": "0.1.2",
     "bignumber.js": "^9.0.0",
     "solidity-coverage": "^0.6.3",
     "truffle": "^5.0.24"
--- a/source/index/package.json
+++ b/source/index/package.json
@@ -19,7 +19,6 @@
   "devDependencies": {
     "@airswap/order-utils": "0.3.3",
     "@airswap/test-utils": "0.1.1",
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -11,13 +11,6 @@
     ethereumjs-util "^6.1.0"
     web3 "^1.0.0-beta.55"
```
poiresel: should tokens@0.1.2 be here as well?

alicevhenshaw: no it's removing it from the lock file. It doesn't need it in `yarn.lock` as it can get version 0.1.2 from the repository itself. It only pulls in airswap modules into here when its not the current version

source/index/package.json
```diff
@@ -19,7 +19,7 @@
   "devDependencies": {
     "@airswap/order-utils": "0.3.3",
     "@airswap/test-utils": "0.1.1",
-    "@airswap/tokens": "0.1.1",
+    "@airswap/tokens": "0.1.2",
```
poiresel: why does index have tokens within it? can it be removed?

poiresel: i could have missed something, just didn't see it.

yarn.lock
alicevhenshaw: just noticed the import of order utils above this. That means the repo is using an old version somewhere too

alicevhenshaw: https://github.com/airswap/airswap-protocols/pull/269/commits/adb413621580f273871414424a67ab7fa1ab3b5e

source/index/package.json
alicevhenshaw: good spot i dont think it needs it

Issue #274, opened 2019-11-07
setRuleAndIntent can have stuck tokens in Delegate


poiresel:
Ensure that unsetRuleAndIntent transfers staking tokens from Indexer to the owner of the delegate

alicevhenshaw:
##Bug explanation

- If `Delegate.setRuleAndIntent` is called once, the Rule is stored in the Delegate, and an Intent to trade is set on the Indexer - including `amountToStake` tokens being `transferFrom` from the delegate owner into the Delegte, and then staked from the Delegate into the Indexer. The Indexer function invoked is `Indexer.setIntent`
- If `Delegate.setRuleAndIntent` is called a _second_ time, with an `amountToStake`, the Delegate assumes that this is the amount of tokens to take from the owner, for the Indexer to then take as stake.
- However, in the second invocation of the function `Indexer.setIntent`, the indexer looks at the difference between the first `amountToStake` and the second `amountToStake`, and either refunds tokens, or takes the difference from the Delegate. This thereby means the Indexer simply transfers the tokens necessary to change the original stake into the updated stake.
- Through this difference in behaviour in the delegate and the indexer, tokens can easily end up stuck in the Delegate.

## Example - increasing stake
**1st stake:** `amountToStake` = 40
Delegate takes 40 from owner, and the Indexer then takes 40 from Delegate in `Indexer.setIntent`
**1st stake:** `amountToStake` = 60
Delegate takes 60 from owner, and the Indexer only takes 20 (the difference) from the Delegate, to end up with 60 tokens total. 40 tokens remain in the Delegate that should be with the owner.

A similar effect occurs when decreasing ones stake.

alicevhenshaw:
This was fixed in https://github.com/airswap/airswap-protocols/pull/277

Issue #273, opened 2019-11-07
Delegate.setRule should take in a Rule struct


poiresel:
Reason: 
Consistency across functions as well as consolidates how many parameters are passed to setRule.

Current impl
```
  function setRule(
    address senderToken,
    address signerToken,
    uint256 maxSenderAmount,
    uint256 priceCoef,
    uint256 priceExp
  ) external onlyOwner {
    _setRule(
      senderToken,
      signerToken,
      maxSenderAmount,
      priceCoef,
      priceExp
    );
  }
```
Proposed change
```
  function setRule(
    address senderToken,
    address signerToken,
    Rule calldata rule
  ) external onlyOwner {
    _setRule(
      senderToken,
      signerToken,
      rule.maxSenderAmount,
      rule.priceCoef,
      rule.priceExp   );
  }
```

ejwessel:
We had this when we had `DelegateManager`.
I think given that it's gone now it's a lot clear of an implementation and definitely warranted to reduce complexity on parameter passing.

alicevhenshaw:
The problem with passing it as a struct is you cannot assign calldata into storage or memory. This means if you have
`setRule(Rule rule)`, you cannot then do
`rules[signer][sender] = rule`
instead you have to assign each element of the struct individually
`rules[signer][sender] = Rule(rule.a, rule.b, rule.c, ...)`
so sadly you dont really get any gains from doing it. Its a limitation of the experimental ABI encoder

poiresel:
Ok, closing

PR #277, opened 2019-11-07
added commit for #273 and fixing unit tests

poiresel:
## Description
In reference to https://github.com/airswap/airswap-protocols/issues/274

Quick description:

- [X] Bug fixes

## Changes

Adding checks in the beginning and end of setRuleAndIntent to handle incrementing or decrementing staked tokens if they already exists

## Tests

36 passing (25s)

## Test Coverage

```
  79 passing (1m)

-----------------------|----------|----------|----------|----------|----------------|
File                   |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
-----------------------|----------|----------|----------|----------|----------------|
 contracts/            |      100 |      100 |      100 |      100 |                |
  Delegate.sol         |      100 |      100 |      100 |      100 |                |
  Imports.sol          |      100 |      100 |      100 |      100 |                |
 contracts/interfaces/ |      100 |      100 |      100 |      100 |                |
  IDelegate.sol        |      100 |      100 |      100 |      100 |                |
-----------------------|----------|----------|----------|----------|----------------|
All files              |      100 |      100 |      100 |      100 |                |
-----------------------|----------|----------|----------|----------|----------------|

Istanbul coverage reports generated

```

### DIFF
--- a/source/delegate/contracts/Delegate.sol
+++ b/source/delegate/contracts/Delegate.sol
@@ -155,7 +155,9 @@ contract Delegate is IDelegate, Ownable {
 
     // get currentAmount staked or 0 if never staked
     uint256 oldStakeAmount = indexer.getStakedAmount(address(this), signerToken, senderToken);
-    if (oldStakeAmount < newStakeAmount) {
+    if (oldStakeAmount == newStakeAmount && oldStakeAmount > 0) {
+      return; // forgo trying to reset intent with non-zero same stake amount
+    } else if (oldStakeAmount < newStakeAmount) {
       // transfer only the difference from the sender to the Delegate.
       require(
         IERC20(indexer.stakingToken())
--- a/source/delegate/contracts/Delegate.sol
+++ b/source/delegate/contracts/Delegate.sol
@@ -132,18 +132,18 @@ contract Delegate is IDelegate, Ownable {
   /**
     * @notice sets a rule on the delegate and an intent on the indexer
     * @dev only callable by owner
-    * @dev delegate needs to be given allowance from msg.sender for the amountToStake
+    * @dev delegate needs to be given allowance from msg.sender for the newStakeAmount
     * @dev swap needs to be given permission to move funds from the delegate
     * @param senderToken address Token the delgeate will send
     * @param senderToken address Token the delegate will receive
     * @param rule Rule Rule to set on a delegate
-    * @param amountToStake uint256 Amount to stake for an intent
+    * @param newStakeAmount uint256 Amount to stake for an intent
     */
   function setRuleAndIntent(
     address senderToken,
     address signerToken,
     Rule calldata rule,
-    uint256 amountToStake
+    uint256 newStakeAmount
   ) external onlyOwner {
     _setRule(
       senderToken,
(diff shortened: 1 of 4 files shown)


source/delegate/contracts/Delegate.sol
```diff
@@ -153,11 +153,13 @@ contract Delegate is IDelegate, Ownable {
       rule.priceExp
     );
 
-    // Transfer the staking tokens from the sender to the Delegate.
-    if (amountToStake > 0) {
+    // get currentAmount staked or 0 if never staked
+    uint256 oldStakingAmount = indexer.getStakedAmount(address(this), signerToken, senderToken);
+    if (oldStakingAmount < amountToStake) {
```
dmosites: Suggest `oldStakeAmount` and `newStakeAmount`

source/delegate/contracts/Delegate.sol
```diff
@@ -167,6 +169,14 @@ contract Delegate is IDelegate, Ownable {
       amountToStake,
       bytes32(uint256(address(this)) << 96) //NOTE: this will pad 0's to the right
     );
+
+    if (oldStakingAmount > amountToStake) {
+      // return excess stake back
+      require(
+        IERC20(indexer.stakingToken())
+        .transfer(msg.sender, oldStakingAmount - amountToStake), "STAKING_TRANSFER_FAILED"
```
dmosites: `STAKING_RETURN_FAILED`

source/delegate/contracts/Delegate.sol
```diff
@@ -167,6 +169,14 @@ contract Delegate is IDelegate, Ownable {
       amountToStake,
       bytes32(uint256(address(this)) << 96) //NOTE: this will pad 0's to the right
     );
+
+    if (oldStakingAmount > amountToStake) {
```
ejwessel: should we place this block above the external `indexer.setIntent()` call?

source/delegate/contracts/Delegate.sol
```diff
@@ -167,6 +169,14 @@ contract Delegate is IDelegate, Ownable {
...
+      // return excess stake back
+      require(
+        IERC20(indexer.stakingToken())
+        .transfer(msg.sender, oldStakingAmount - amountToStake), "STAKING_TRANSFER_FAILED"
+      );
+    }
```
ejwessel: we have the conditions for:
```
if (oldStakingAmount < amountToStake)
if (oldStakingAmount > amountToStake)
```
but not when `oldStakingAmount == amountToStake` just to confirm, if it's equal we're assuming they're likely updating the rule and we probably don't need to reset the intent.

poiresel: yes

source/delegate/contracts/Delegate.sol
poiresel: we can't because the staked amount from the indexer would not be in this contract yet.

source/delegate/test/Delegate.js
```diff
@@ -215,6 +378,72 @@ contract('Delegate Integration Tests', async accounts => {
...
+      let scoreAfter = await indexer.getStakedAmount(
+        aliceDelegate.address,
+        tokenDAI.address,
+        tokenWETH.address
+      )
+      equal(scoreAfter.toNumber(), 0, 'intent score is incorrect')
+
+      //check owner stake balance has been reduced by 0
+      let stakingTokenBal = await stakingToken.balanceOf(aliceAddress)
+      console.log()
```
ejwessel: is this console log needed?

source/delegate/contracts/Delegate.sol
ejwessel: could we forego setting the intent then at the end?

source/delegate/contracts/Delegate.sol
ejwessel: of course, thanks for the clarification.

source/delegate/test/Delegate.js
```diff
@@ -165,13 +175,166 @@ contract('Delegate Integration Tests', async accounts => {
...
+      )
+      equal(
+        scoreAfter.toNumber(),
+        2 * INTENT_AMOUNT,
+        'intent score is incorrect'
+      )
+
+      //check owner stake balance has been reduced
+      let stakingTokenBal = await stakingToken.balanceOf(aliceAddress)
+      console.log()
```
alicevhenshaw: console.log?

source/delegate/test/Delegate.js
```diff
@@ -165,13 +175,166 @@ contract('Delegate Integration Tests', async accounts => {
...
+      let scoreAfter = await indexer.getStakedAmount(
+        aliceDelegate.address,
+        tokenDAI.address,
+        tokenWETH.address
+      )
+      equal(scoreAfter.toNumber(), 0, 'intent score is incorrect')
+
+      //check owner stake balance has been reduced
+      let stakingTokenBal = await stakingToken.balanceOf(aliceAddress)
+      console.log()
```
alicevhenshaw: and another

source/delegate/contracts/Delegate.sol
poiresel: so i'd prefer not having the delegate reach in and check if an index exists and whether an intent exists. I'll put in a check to forego at zero but any further checks to the internals of index doesn't see right.

source/delegate/contracts/Delegate.sol
```diff
@@ -155,7 +155,9 @@ contract Delegate is IDelegate, Ownable {
 
     // get currentAmount staked or 0 if never staked
     uint256 oldStakeAmount = indexer.getStakedAmount(address(this), signerToken, senderToken);
-    if (oldStakeAmount < newStakeAmount) {
+    if (oldStakeAmount == newStakeAmount && oldStakeAmount > 0) {
```
ejwessel: good catch on the ` && oldStakeAmount > 0`

PR #278, opened 2019-11-07
Order validation

alicevhenshaw:
## Description

A new function `checkOrder` to check whether an order is likely to pass if submitted to Swap. It checks party's balances, allowances, signature, nonces and expiry. It cannot check the details of authorising a sender or empty signatures as it cannot tell who is going to submit the order to the swap contract.


utils/order-utils/src/signatures.js
```diff
@@ -91,14 +95,32 @@ module.exports = {
...
       signatory: EMPTY_ADDRESS,
-      validator: EMPTY_ADDRESS,
+      validator: verifyingContract,
       version: signatures.INTENDED_VALIDATOR,
       v: '0',
       r: '0x0',
       s: '0x0',
     }
   },
+  async isSignatureValid(order) {
```
cloudonshore: I don't believe the `async` is needed here since there is no `await` in the function body

utils/order-utils/src/signatures.js
```diff
@@ -16,6 +16,10 @@
 
 const ethUtil = require('ethereumjs-util')
 const sigUtil = require('eth-sig-util')
+const web3Eth = require('web3-eth')
+
+const eth = new web3Eth('http://127.0.0.1:8545')
```
cloudonshore: I'd like to be able to include this module without having an ethereum node running locally. I'm concerned an exception would be thrown in that case. Maybe this could be driven by a env variable so that if the env variable isn't present it doesn't try connecting to a local node?

package.json
```diff
@@ -12,7 +12,7 @@
     "compile": "lerna run compile",
     "hint": "yarn solhint \"./source/**/*.sol\"",
     "lint": "yarn eslint \"./**/*.js\"",
-    "ganache": "ganache-cli -p 8545 --gasLimit 0xfffffffffff --time '2017-05-10T00:00:00+00:00'",
+    "ganache": "ganache-cli -p 8545 --gasLimit 0xfffffffffff --time '2017-05-10T00:00:00+00:00' -s 0",
```
ejwessel: why do we need the seed flag?

poiresel: put it in as needed to have a fixed account address in ganache for testing.

utils/order-utils/src/signatures.js
poiresel: maybe use an infura key in the .env if the network is rinkeby/mainnet then? as we're making contract lookups?

utils/order-utils/test/orders.js
```diff
@@ -1,14 +1,21 @@
 var expect = require('chai').expect
+const assert = require('assert')
 
 const { orders } = require('@airswap/order-utils')
 
 describe('Orders', async () => {
   const senderWallet = '0xbabe31056c0fe1b704d811b2405f6e9f5ae5e59d'
   const signerWallet = '0x9d2fb0bcc90c6f3fa3a98d2c760623a4f6ee59b4'
+  const knownGanacheWallet = '0xe092b1fa25DF5786D151246E492Eed3d15EA4dAA'
```
poiresel: maker this is accounts[0] when seed == 0 for ganache

utils/order-utils/test/orders.js
```diff
@@ -1,14 +1,21 @@
...
 
   const ASTAddress = '0xcc1cbd4f67cceb7c001bd4adf98451237a193ff8'
```
poiresel: mark these are the rinkeby addresses

yarn.lock
```diff
@@ -3917,7 +4000,7 @@ ethers@4.0.0-beta.3:
     uuid "2.0.1"
     xmlhttprequest "1.8.0"
 
-ethers@^4.0.27:
+ethers@^4.0.27, ethers@^4.0.38:
```
poiresel: do we need to use both versions?

package.json
alicevhenshaw: Essentially you can't pull in `accounts[0]` to the order tests as it isnt a truffle file. Instead I have just copied in the address. It's needed so that we can test ganache signing being provided only the address.

utils/order-utils/src/signatures.js
alicevhenshaw: Removed this as the corresponding await

PR #282, opened 2019-11-12
Add slither

poiresel:
## Description

 To use:
- Create a Python 3.6 `virtualenv --python=python3.6 venv`
- Activate the virtualenv `source venv/bin/activate`
- Install slither `pip install slither-analyzer`
- From root directory run `./scripts/slither.sh`

- [X] N/A


source/delegate/reports/SPEC.md
```diff
@@ -0,0 +1,107 @@
+# Report: Delegate
```
ejwessel: is this file necessary?

poiresel: removing its part of the delegate-report branch

Issue #280, opened 2019-11-08
Remove signing logic from and rename "getOrder" call (order-utils)


dmosites:
Would like to simplify the `getOrder` function and consider renaming it to `createOrder` or similar.

alicevhenshaw:
I agree. Creating the order, and getting the signature should be 2 different things. One you do using orders.js, the other you do using signatures.js. It would lead to a nice separation of concerns and decrease the code complexity

ejwessel:
implemented in #284

PR #284, opened 2019-11-12
Separate signature from getOrder

alicevhenshaw:
## Description

https://github.com/airswap/airswap-protocols/issues/280
Separating creation of orders from creation of signatures in the `order-utils` package.

## Changes

- Removal of getWeb3Signature from `orders.getOrder`
- getWeb3Signature now takes its a provider as a parameter
- Updating tests to add their own signature where needed

## Tests
All passing

## Test Coverage
Not changed


alicevhenshaw:
@poiresel have changed known ganache address to be with all caps. However in all our other tests when we have AST address, etc, its always in camel case and not in capitals - so have left those.

PR #286, opened 2019-11-15
Add optimizer to truffle-config + update truffle plugins

poiresel:
## Description

Quick description:
- [X] New features

## Changes

Two quick changes
1. Update the truffle-verify + truffle-flatten to the most recent versions
2. Add the optimizer to the truffle-config.js, Test were run using https://github.com/cgewecke/eth-gas-reporter but not significant reductions were seen. Will look to benchmark in test and mainnet deploys though.

## Tests

N/A

## Test Coverage

N/A


package.json
```diff
@@ -33,8 +33,8 @@
     "solhint": "^2.1.0",
     "truffle": "^5.0.32",
     "ganache-time-traveler": "^1.0.5",
-    "truffle-flatten": "^1.0.5",
-    "truffle-verify": "^1.0.3",
+    "truffle-flatten": "^1.0.6",
+    "truffle-verify": "^1.0.4",
```
alicevhenshaw: shall we remove hats? to fix the version and increase it manually?

poiresel: can do

source/delegate-factory/truffle-config.js
```diff
@@ -30,6 +30,12 @@ module.exports = {
   compilers: {
     solc: {
       version: '0.5.12',
+      settings: {
+        optimizer: {
+          enabled: true,
+          runs: 20000,
+        },
+      },
       optimization: false,
```
alicevhenshaw: Should this read `optimization: false`?

alicevhenshaw: I see youve now removed this anyway

alicevhenshaw:
Should this be added for `delegate` and `index` too?

package.json
poiresel: i don't think so, did the upgrade for visibility but based on tests yarn will upgrade if needed.

PR #289, opened 2019-11-15
Order check nft

alicevhenshaw:
## Description

Added checks to `checkOrder` for swaps that include ERC721s

### DIFF
--- a/utils/order-utils/package.json
+++ b/utils/order-utils/package.json
@@ -19,8 +19,8 @@
     "@airswap/tokens": "0.1.2",
     "eth-sig-util": "^2.2.0",
     "ethereumjs-util": "^6.1.0",
-    "ethers": "^4.0.27",
-    "web3-eth": "1.2.2",
-    "web3": "^1.0.0-beta.55"	
+    "ethers": "^5.0.0-beta.159",
+    "web3": "^1.0.0-beta.55",
+    "web3-eth": "1.2.2"
   }
 }
--- a/utils/order-utils/package.json
+++ b/utils/order-utils/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/airswap/airswap-protocols"
   },
   "scripts": {
-    "test": "mocha test"
+    "test": "mocha test --timeout 3000"
   },
   "devDependencies": {
     "chai": "^4.2.0",
--- a/utils/order-utils/package.json
+++ b/utils/order-utils/package.json
@@ -8,6 +8,7 @@
     "url": "https://github.com/airswap/airswap-protocols"
   },
   "scripts": {
+    "ganache": "ganache-cli -p 8545 --gasLimit 0xfffffffffff --time '2017-05-10T00:00:00+00:00' -s 0",
     "test": "mocha test --timeout 3000"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


utils/order-utils/package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://github.com/airswap/airswap-protocols"
   },
   "scripts": {
-    "test": "mocha test"
+    "test": "mocha test --timeout 3000"
```
poiresel: why is the timeout required?

utils/order-utils/test/orders.js
```diff
@@ -168,6 +172,122 @@ describe('Orders', async () => {
...
+    )
+
+    const errors = await orders.checkOrder(order, 'rinkeby')
+    assert.equal(errors.length, 1)
+    assert.equal(errors[0], 'sender no NFT approval')
+  })
+
+  it('Check NFT order to a contract', async () => {
+    const order = await orders.getOrder({
+      expiry: '1604787494',
```
ejwessel: can we capture all instances of `1604787494` and put it at the top as a const so that we know it's reference to the ganache time `2017-05-10T00:00:00+00:00`?

utils/order-utils/package.json
alicevhenshaw: The tests are quite time consuming as they query rinkeby. Normally they run in around 800ms-1400ms. The Mocha default is 2000ms cut off, so I increased this just in case something runs inefficiently.

utils/order-utils/test/orders.js
alicevhenshaw: Its actually `Saturday, 7 November 2020 22:18:14`
But yes will update accordingly

alicevhenshaw: This is one of my laptop bookmarks lol
https://www.epochconverter.com/

utils/order-utils/test/orders.js
```diff
@@ -27,12 +27,13 @@ describe('Orders', async () => {
...
   const NOV_7_2020_22_18_14 = '1604787494'
+  const MAY_11_2017_00_00_00 = '1494460800'
 
   orders.setVerifyingContract(rinkebySwap)
 
   it('Checks that a generated order is valid', async () => {
     const order = await orders.getOrder({
-      expiry: '1494460800',
+      expiry: MAY_11_2017_00_00_00,
       nonce: '101',
```
poiresel: can nonce be pulled out as well to be a const?

alicevhenshaw: Yes, very happy to. Just hoping for a bit more discussion around this as I feel like its an undefined rule of when we pull out a const and when we don't.

I see that pulling out the timestamp above provides way more clarity on the variable - what was previously a random looking string of numbers is now a clearly defined date and so it can be seen that it is may 11th etc. Similar clarity is gained by pulling out an address and replacing it with `ASTAddress`.

I'm just wondering if pulling out nonce gives any more clarity than it already has as is? Or is the reason you're asking for this is because the same number is used many times? And should I also be pulling out the `param: x` numbers as const too?

utils/order-utils/package.json
ejwessel: May want to disregard this since there is no `truffle-config.js` in this directory since it's not connecting through that method.

alicevhenshaw:
@ejwessel 
1. Good point, I'll add the ganache script to the repo. I would say that its not entirely unexpected that it cannot run without ganache running though, that is true of all of the sub repos.
2. This address is a non-ownable smart contract. It merely exists on rinkeby. You can view it here https://rinkeby.etherscan.io/address/0xf727956c4cfd20b9c8d463218a65e751891da3e6#code

Issue #298, opened 2019-11-19
QSP-6: Unchecked Return Value


alicevhenshaw:
## Description from Quantstamp

**Contract(s) affected:** Wrapper.sol Swap.sol
**Description:** Most functions will return a true or false value upon success. Some functions, like send(), are more crucial to check than others. It's important to ensure that every necessary function is checked.
On L151 of Wrapper, the external call.value() is not checked for success, which may cause ether transfers to the user to fail. A similar issue exists for transfer() on L127 of Wrapper, and L340 of Swap.
**Recommendation**: The external call.value() result should be checked for success by changing the line to: 
`(bool success, ) =msg.sender.call.value(order.signer.param)("")`
followed by a check on success.
Additionally, on L127, the return value of Wrapper.transfer() and L340 of Swap.transferToken() should be checked

## Response from Airswap

`Wrapper msg.sender.call.value:` We will be updating the contract with the recommended fix.

`Wrapper L#127 WETH.transfer():` The WETH contract we use is a fixed address at `0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2`, which can be viewed and is verified on Etherscan. The transfer and transferFrom functions never return false. If something goes wrong in the function call, the contract throws instead of returning false.

`Swap L#340 IERC721(token).transferFrom():` IERC721 function transferFrom does not return a boolean indicating the success or failure of the function. We therefore cannot check whether

PR #301, opened 2019-11-19
Revert if ETH return is unsuccessful

alicevhenshaw:
## Description

https://github.com/airswap/airswap-protocols/issues/298

Quick description:

- [ ] N/A
- [ ] New features
- [X] Bug fixes
- [ ] Typo/small tweaks

## Changes

- Added a revert statement when ETH fails to send to a non-payable contract in the wrapper
- Created a mock contract that is non-payable to test this new revert
- Added an integration test that uses the new mock to test the revert

## Tests
```
Wrapper:
  39 passing (10s)
```

### DIFF
--- a/source/wrapper/contracts/HelperMock.sol
+++ b/source/wrapper/contracts/HelperMock.sol
@@ -1,0 +1,31 @@
+pragma solidity 0.5.12;
+pragma experimental ABIEncoderV2;
+
+import "@airswap/types/contracts/Types.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "./Wrapper.sol";
+
+contract HelperMock {
+
+  Wrapper public wrapper;
+
+  constructor(Wrapper helperWrapper) public {
+    wrapper = helperWrapper;
+  }
+
+  function forwardSwap(
+    Types.Order calldata order
+  ) external {
+    wrapper.swap(order);
+  }
+
+  function authorizeWrapperToSend() external {
+    ISwap swap = wrapper.swapContract();
+    swap.authorizeSender(address(wrapper));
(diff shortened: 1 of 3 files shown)


source/wrapper/contracts/Wrapper.sol
```diff
@@ -148,7 +148,9 @@ contract Wrapper is Ownable {
 
       // Transfer ether to the user.
       // solium-disable-next-line security/no-call-value
```
poiresel: can remove the  // solium-disable-next-line security/no-call-value now

alicevhenshaw: It's still doing a call value, and so is still considered a security flaw as you pass the recipient contract access to all gas and allow reentrancy. Solium will still warn of this even if we catch the bool as you can do reentrancy as then succeed.

poiresel: Interesting, ok.

source/wrapper/contracts/HelperMock.sol
```diff
@@ -0,0 +1,31 @@
...
+  function authorizeWrapperToSend() external {
+    ISwap swap = wrapper.swapContract();
+    swap.authorizeSender(address(wrapper));
+  }
+
+  function approveToken(IERC20 token, address toApprove, uint256 amount) external {
+    token.approve(toApprove, amount);
+  }
+
+}
```
ejwessel: Can we add a note somewhere in here to denote its use is exclusively for non payable?

alicevhenshaw: Yeah nice, I'll add a natspec at the top of the contract to explain why its been created.

Issue #294, opened 2019-11-19
QSP-7 Gas Usage / for Loop Concerns - Swap.cancel()


ejwessel:
## Description from Quantstamp:

Contract(s) affected: Swap.sol, Index.sol
Description: Gas usage is a main concern for smart contract developers and users, since high gas costs may prevent users from wanting to use the smart contract. Even worse, some gas usage issues may prevent the contract from providing services entirely. For example, if a for loop requires too much gas to exit, then it may prevent the contract from functioning correctly entirely. It is best to break such loops into individual functions as possible. In particular, the Swap.cancel() function may fail if the array of nonces is too long...
Recommendation: Although the user could re-invoke the Swap.cancel() function by breaking the array up across multiple transactions, we recommend adding a comment to the function's description to indicate such potential issues.

## Response from Airswap:
Will we be fixing this issue?
Yes, we will add a comment to the function's description to indicate potential issues and how to remedy if the array of nonces is too long.

alicevhenshaw:
So my maths on Swap.cancel having analysed cancelling different length arrays of nonces:
- Function overhead: 22201
- Cost to cancel a nonce: 22831
- Therefore cost to cancel 100 nonces: 2,305,301
- Total nonces you could cancel in 1 transaction on mainnet before hitting the block gas limit: 437

alicevhenshaw:
Explanation added to comments in contract https://github.com/airswap/airswap-protocols/pull/306

ejwessel:
Analysis here: https://github.com/airswap/airswap-protocols/pull/315

PR #306, opened 2019-11-20
QSP-7 Gas Usage / for Loop Concerns - Swap.cancel()

ejwessel:
## Description
https://github.com/airswap/airswap-protocols/issues/294
Update the `Swap.cancel()` to properly indicate issues with gas usage and how to remedy.

- [ ] N/A
- [ ] New features
- [ ] Bug fixes
- [X] Typo/small tweaks

## Changes
Update @dev with description of gas usage.

## Tests
N/A

### DIFF
--- a/source/swap/contracts/Swap.sol
+++ b/source/swap/contracts/Swap.sol
@@ -174,8 +174,8 @@ contract Swap is ISwap {
     * @dev Emits a Cancel event
     * @dev If the list of nonces to cancel is too long, 
       there is a potential issue of out of gas error. 
-      To remedy this it is best to break the list into 
-      smaller chunks and submit the cancel() this way instead
+      It is best to break the list into smaller chunks 
+      and submit the cancel() this way instead
     * @param nonces uint256[] List of nonces to cancel
     */
   function cancel(
--- a/source/swap/contracts/Swap.sol
+++ b/source/swap/contracts/Swap.sol
@@ -172,10 +172,7 @@ contract Swap is ISwap {
     * @notice Cancel one or more open orders by nonce
     * @dev Cancelled nonces are marked UNAVAILABLE (0x01)
     * @dev Emits a Cancel event
-    * @dev If the list of nonces to cancel is too long, 
-    * there is a potential issue of out of gas error. 
-    * It is best to break the list into smaller chunks 
-    * and submit the cancel() this way instead.
+    * @dev Out of gas may occur on arrays of length > 400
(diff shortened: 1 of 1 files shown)


source/swap/contracts/Swap.sol
```diff
@@ -172,6 +172,10 @@ contract Swap is ISwap {
     * @notice Cancel one or more open orders by nonce
     * @dev Cancelled nonces are marked UNAVAILABLE (0x01)
     * @dev Emits a Cancel event
+    * @dev If the list of nonces to cancel is too long, 
+      there is a potential issue of out of gas error.
```
alicevhenshaw: Add a star at the beginning of each line

alicevhenshaw:
I personally don't think we need to explain ourselves here.

My maths on `Swap.cancel` having analysed cancelling different length arrays of nonces:
- Function overhead: 22201
- Cost to cancel a nonce: 22831

Therefore cost to cancel 100 nonces: 2,305,301
Total nonces you could cancel in 1 transaction on mainnet before hitting the block gas limit: 437

If someone is trying to cancel even 100 nonces I’d be surprised, let alone 437. They can use the function `invalidate` to set a minimum nonce.

ejwessel:
I agree with you that it may be an overly informative to explain ourselves. However, do you think it reflects better on us to denote that we're aware about the limitations and have vocalized it? I'm happy to close this either way.

alicevhenshaw:
Currently it's a long explanation that would make people worry it actually might happen to them, without saying what the limits of the function really are. Imo if its going to be included you should keep it brief and include actual data - I'd rather have a 1 liner that just says:
`* @dev Out of gas may occur on arrays of length > 400`
This way the reader gets more data on when the error occurs.

ejwessel:
I'm all for keeping a concise message. 

I'll wait for input from others on where we should go from here.

poiresel:
* @dev Out of gas may occur on arrays of length > 400 seems like a nice informative succinct message.

source/swap/contracts/Swap.sol
```diff
@@ -172,6 +172,7 @@ contract Swap is ISwap {
     * @notice Cancel one or more open orders by nonce
     * @dev Cancelled nonces are marked UNAVAILABLE (0x01)
     * @dev Emits a Cancel event
+    * @dev Out of gas may occur on arrays of length > 400
```
poiresel: occur in arrays versus occur on arrays

PR #310, opened 2019-11-20
Remove throwing events for non-state changes in Swap and Delegate contract

poiresel:
## Description

Quick description:
- [X] Typo/small tweaks

## Changes

Put in checks to ensure a state change was occurring before emitting events. Added some tests to ensure calling the functions duplicate times would not emit an event.

## Tests
```
  73 passing (15s)

lerna success run Ran npm script 'test' in 8 packages in 181.2s:
lerna success - @airswap/delegate-factory
lerna success - @airswap/delegate
lerna success - @airswap/index
lerna success - @airswap/indexer
lerna success - @airswap/swap
lerna success - @airswap/types
lerna success - @airswap/wrapper
lerna success - @airswap/order-utils
Done in 387.51s.


## Test Coverage

Still 100%
```

### DIFF
--- a/source/delegate/contracts/Delegate.sol
+++ b/source/delegate/contracts/Delegate.sol
@@ -415,13 +415,15 @@ contract Delegate is IDelegate, Ownable {
     address signerToken
   ) internal {
 
-    // Delete the rule.
-    delete rules[senderToken][signerToken];
-
-    emit UnsetRule(
-      owner(),
-      senderToken,
-      signerToken
+    // using non-zero rule.priceCoef for rule existence check
+    if (rules[senderToken][signerToken].priceCoef > 0) {
+      // Delete the rule.
+      delete rules[senderToken][signerToken];
+      emit UnsetRule(
+        owner(),
+        senderToken,
+        signerToken
     );
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


source/delegate/contracts/Delegate.sol
```diff
@@ -415,13 +415,15 @@ contract Delegate is IDelegate, Ownable {
...
   ) internal {
 
-    // Delete the rule.
-    delete rules[senderToken][signerToken];
-
-    emit UnsetRule(
-      owner(),
-      senderToken,
-      signerToken
+    // using non-zero rule.priceCoef for rule existence check
```
alicevhenshaw: Do you think we should add a require on _setRule for
`require(priceCoef > 0)` ?

poiresel: Yes, if someone wants to do a priceCoef == 0, they honestly should not use the Delegate to do this.

source/delegate/contracts/Delegate.sol
```diff
@@ -388,6 +388,7 @@ contract Delegate is IDelegate, Ownable {
     uint256 priceCoef,
     uint256 priceExp
   ) internal {
+    require(priceCoef > 0, "REQUIRE_NONZERO_PRICE");
```
alicevhenshaw: We don't normally put the word `REQUIRE` in the actual require messages. Maybe something like `INVALID_PRICE_COEF` ?

PR #312, opened 2019-11-21
Adding a note for testing Delegate contracts

dmosites:
## Description

Quick description: Adding a note for testing Delegate contracts

- [ ] N/A
- [ ] New features
- [ ] Bug fixes
- [X] Typo/small tweaks

## Changes

Non-functional change

### DIFF
--- a/source/delegate/README.md
+++ b/source/delegate/README.md
@@ -38,3 +38,5 @@ Run an instance of `ganache-cli` before running tests.
 ```
 yarn ganache
 ```
+
+Delegate tests require build artifacts from `@airswap/swap` and `@airswap/tokens`. Before running tests, run `yarn compile` in both the [swap](../swap) and [tokens](../tokens) packages in this repository.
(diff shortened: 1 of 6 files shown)


poiresel:
Add a similar note for order-utils as well.

source/delegate/README.md
```diff
@@ -38,3 +38,5 @@ Run an instance of `ganache-cli` before running tests.
 ```
 yarn ganache
 ```
+
+Delegate tests require build artifacts from `@airswap/swap` and `@airswap/tokens`. Before running tests, run `yarn compile` in both the [swap](../swap) and [tokens](../tokens) packages in this repository.
```
ejwessel: While this is true. Why not instruct them to run `yarn compile` from the root directory?

dmosites: Yep turns out `order-utils` depends on builds so everything should be compiled prior to running tests on an individual package.

PR #317, opened 2019-11-22
Update ERC20_INTERFACE_ID in Delegate and constants.js and upgraded o…

poiresel:
## Description

Quick description:
- [X] Typo/small tweaks

## Changes

Changed ERC20_INTERFACE_ID bytes4 to include function approve and totalSupply which update the hash to 0x36372b07.
Only Solidity contract that used ERC20_INTERFACE_ID was Delegate which was updated as well.
Updated order-utils version as the constants.js needed to be updated since all tests rely on this.

The new hash represents:
bytes4(keccak256('transfer(address,uint256)')) ^ bytes4(keccak256('transferFrom(address,address,uint256)')) ^
bytes4(keccak256('balanceOf(address)')) ^ bytes4(keccak256('allowance(address,address)'))
    ^ bytes4(keccak256('approve(address,uint256)')) ^ bytes4(keccak256('totalSupply'))

Only other usage of this ID can be see:
https://github.com/0xcert/framework/blob/master/packages/0xcert-ethereum-erc20-contracts/src/contracts/token.sol

## Tests

All pass

## Test Coverage

Just a rename, and everything still covered.

### DIFF
--- a/utils/order-utils/test/orders.js
+++ b/utils/order-utils/test/orders.js
@@ -146,7 +146,7 @@ describe('Orders', async () => {
       GANACHE_PROVIDER
     )
 
-    order.signature.v += 1
+    order.signature.v -= 1
     const errors = await orders.checkOrder(order, 'rinkeby')
     assert.equal(errors.length, 2)
     assert.equal(errors[1], 'Signature invalid')
--- a/source/delegate-factory/package.json
+++ b/source/delegate-factory/package.json
@@ -21,7 +21,7 @@
     "verify": "truffle run verify"
   },
   "devDependencies": {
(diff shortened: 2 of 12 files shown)


utils/order-utils/test/orders.js
```diff
@@ -146,7 +146,7 @@ describe('Orders', async () => {
       GANACHE_PROVIDER
     )
 
-    order.signature.v += 1
+    order.signature.v -= 1
```
poiresel: This was added since the order.signature.v is now 28 and adding one would take it out of the range, so instead subtracted 1.

alicevhenshaw: Nice catch

PR #320, opened 2019-11-25
Rinkeby Update Deployed Addresses

ejwessel:
## Description
https://www.notion.so/fluidity/Rinkeby-Deploy-2019-11-25-d442fb6faffd4eaf82b72166a907441c


poiresel:
Delegate pulls in IIndexer.sol but does not have it as a dependency in the package.json

PR #321, opened 2019-11-25
Updating return type of ISwap functions

alicevhenshaw:
## Description

ISwap had an incorrect return type for 2 functions. This is a remnant from when we have expiries.

Quick description:

- [ ] N/A
- [ ] New features
- [X] Bug fixes
- [ ] Typo/small tweaks

### DIFF
--- a/source/delegate/package.json
+++ b/source/delegate/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@airswap/delegate",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "description": "Manages delegated trading rules for use in the Swap Protocol",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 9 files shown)


source/delegate/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@airswap/indexer": "0.4.2",
-    "@airswap/swap": "2.3.2",
+    "@airswap/swap": "2.3.3",
```
poiresel: do we need to increment delegate as its dependency has updated.

source/wrapper/package.json
```diff
@@ -27,7 +27,7 @@
     "solidity-coverage": "^0.6.3"
   },
   "dependencies": {
-    "@airswap/swap": "2.3.2",
+    "@airswap/swap": "2.3.3",
```
poiresel: increment wrapper as well as its dependency was upgraded.

source/delegate/package.json
alicevhenshaw: That's not how hwe've done it in previous PRs. We normally just bump the thing that's changed.
e.g. Ethan's PR this morning https://github.com/airswap/airswap-protocols/pull/319/files

I do see your point though.

alicevhenshaw: Bumped delegate and therefore delegate factory

source/wrapper/package.json
alicevhenshaw: Bumped wrapper

PR #322, opened 2019-11-25
Converting buffer to hex for ecsign signatures

dmosites:
## Description

`ecsign` was returning Buffer instead of hex.

- [ ] N/A
- [ ] New features
- [X] Bug fixes
- [ ] Typo/small tweaks


dmosites:
> does a similar behavior need to be in getWeb3Signature?

No because `ethUtil.fromRpcSig` does the conversion.

PR #327, opened 2019-11-26
always cast expiry, nonce, v to string

perich:
- Cast all values to string
- Lowercase all addresses

### DIFF
--- a/source/wrapper/test/Wrapper.js
+++ b/source/wrapper/test/Wrapper.js
@@ -129,7 +129,12 @@ contract('Wrapper', async ([aliceAddress, bobAddress, carolAddress]) => {
       result = await getResult(swapContract, result.tx)
       emitted(result, 'Swap')
 
-      ok(await balances(wrapperAddress, [[tokenDAI, 0], [tokenWETH, 0]]))
+      ok(
+        await balances(wrapperAddress, [
+          [tokenDAI, 0],
+          [tokenWETH, 0],
+        ])
+      )
       ok(await balances(aliceAddress, [[tokenWETH, 10]]))
       ok(await balances(bobAddress, [[tokenDAI, 50]]))
     })
@@ -187,7 +192,12 @@ contract('Wrapper', async ([aliceAddress, bobAddress, carolAddress]) => {
       result = await getResult(swapContract, result.tx)
       emitted(result, 'Swap')
 
-      ok(await balances(wrapperAddress, [[tokenDAI, 0], [tokenWETH, 0]]))
+      ok(
+        await balances(wrapperAddress, [
+          [tokenDAI, 0],
+          [tokenWETH, 0],
+        ])
+      )
       ok(await balances(aliceAddress, [[tokenDAI, 850]]))
     })
   })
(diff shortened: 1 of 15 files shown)


yarn.lock
```diff
@@ -9677,12 +9684,13 @@ truffle-assertions@^0.9.1:
     assertion-error "^1.1.0"
     lodash.isequal "^4.5.0"
 
-truffle-flatten@^1.0.5:
-  version "1.0.5"
-  resolved "https://registry.yarnpkg.com/truffle-flatten/-/truffle-flatten-1.0.5.tgz#0d39a73ae689b78cec8764bf09a9fe0dbd35f1ca"
-  integrity sha512-oGCmrj0h/Dpr5K2TDOEb7BQU5NG8pgDv0VLmPTL2+Il8Cd+BzgaT6E7+NpM52YeyCvPutYp5sAaEetDJqN1n/w==
+truffle-flatten@^1.0.6:
```
alicevhenshaw: Do you mind reverting the yarn.lock to how it was before please? This includes an update to 2 truffle tools we use and we should assess whether we want the update on those before it goes in.

perich: it was updated 11 days ago in package.json to 1.0.6, but the lockfile was never bumped. they should be reconciled

https://github.com/airswap/airswap-protocols/commit/2694f0a44400e5e6eaf723cce7a489240c22d200

utils/order-utils/src/orders.js
```diff
@@ -288,6 +288,18 @@ const isValidOrder = order => {
   )
 }
 
+function lowerCaseAddresses(order) {
+  for (var key in order) {
+    if (typeof order[key] === 'object') {
+      lowerCaseAddresses(order[key])
+    }
+    if (typeof order[key] === 'string' && order[key].indexOf('0x') === 0) {
```
alicevhenshaw: This line will catch `kind`, `version`, `r` and `s` as well as addresses. I imagine that's fine, but just checking you have noticed that?

dmosites: Yes.

yarn.lock
alicevhenshaw: Ah good spot hadn't noticed that.

perich: if we want to use 1.0.6, yarn.lock should be bumped since it's gonna get bumped on any machine that runs `yarn` in the future, and I think also in CI.

if don't want to use 1.0.6, we should revert the commit linked above. 


But, it shouldn't stay as is since it's a mismatch in the current state.

Let me know what you think and I can update @alicevhenshaw @poiresel

source/wrapper/test/Wrapper.js
```diff
@@ -129,7 +129,12 @@ contract('Wrapper', async ([aliceAddress, bobAddress, carolAddress]) => {
       result = await getResult(swapContract, result.tx)
       emitted(result, 'Swap')
 
-      ok(await balances(wrapperAddress, [[tokenDAI, 0], [tokenWETH, 0]]))
+      ok(
```
alicevhenshaw: if we're pushing these yarn lint changes to, for example, `wrapper`. Does `@airswap/wrapper` need bumping?

Issue #323, opened 2019-11-25
Remove msg.sender check in Delegate.providerOrder


poiresel:
Current Behavior:
```
require(order.signer.wallet == msg.sender,
      "SIGNER_MUST_BE_SENDER");
```
New Behavior:
```
    require(order.signature.v != 0,
      "SIGNATURE_MUST_BE_SENT");
```

Purpose:
This change unlocks the ability for a Delegate contract interact with a Wrapper contract to remove the usage of WETH from an EOA wallet.

Because a signature is ultimately checked on the signer as long as the signature is not empty in the swap contract as well as the signer has not signer authorized the delegate, there is still a signing check taking place further down the transaction line.

alicevhenshaw:
> "Because a signature is ultimately checked on the signer"

What do you mean by this?

poiresel:
This was done in #334.

PR #334, opened 2019-11-26
Delegate signature

alicevhenshaw:
## Description

Updating the Delegate to require a signature on orders. Per issue https://github.com/airswap/airswap-protocols/issues/323

Quick description:

- [X] Typo/small tweaks

## Changes

- Updated the require statement in the delegate
- Added signatures to tests that previously didnt provide them
- Added 2 new tests

## Tests

Delegate: `  93 passing (17s)`

## Test Coverage

Unchanged


source/delegate/contracts/Delegate.sol
```diff
@@ -219,8 +219,8 @@ contract Delegate is IDelegate, Ownable {
 
     Rule memory rule = rules[order.sender.token][order.signer.token];
 
-    require(order.signer.wallet == msg.sender,
-      "SIGNER_MUST_BE_SENDER");
+    require(order.signature.v != 0,
+      "SIGNATURE_MUST_BE_SENT");
```
poiresel: SIGNATURE_MUST_BE_SET?

source/delegate/test/Delegate.js
```diff
@@ -910,8 +949,8 @@ contract('Delegate Integration Tests', async accounts => {
 
       // Succeeds on the Delegate, fails on the Swap.
```
poiresel: Message doesn't match case "Fails on Delegate due to not sending signature"

source/delegate/contracts/Delegate.sol
alicevhenshaw: We decided to keep this consistent with the wrapper and leave it as `SENT`.

source/delegate/test/Delegate.js
```diff
@@ -1069,6 +1148,41 @@ contract('Delegate Integration Tests', async accounts => {
       )
     })
 
+    it('Gets a quote to sell 1 WETH where sender != signer, passes', async () => {
```
ejwessel: did you want to add a test before this that tests the failure of having no signature?

alicevhenshaw: Its on line 953

PR #362, opened 2019-12-11
Dependency revamp

alicevhenshaw:
## Description

- `order-utils` uses `ethereumjs-abi` which was previously a devDependency. The only reason we weren't hitting a problem of it being imported was because `eth-sig-util` imports it. Considering we use a carat for eth-sig I dont think we should rely on it being there forever. Have moved this dependency into `order-utils` and made it a `dependency`
- most repos were importing `solidity-coverage` independently. I moved it to be a global `devDependency`, seeing as this is where we import all of our other test-related packages.

### DIFF
--- a/source/delegate-factory/package.json
+++ b/source/delegate-factory/package.json
@@ -27,7 +27,7 @@
     "verify": "truffle run verify"
   },
   "devDependencies": {
-    "@airswap/order-utils": "0.3.13",
+    "@airswap/order-utils": "0.3.14",
     "@airswap/test-utils": "0.1.3"
   },
   "dependencies": {
--- a/source/delegate-factory/package.json
+++ b/source/delegate-factory/package.json
@@ -28,8 +28,7 @@
   },
   "devDependencies": {
     "@airswap/order-utils": "0.3.13",
-    "@airswap/test-utils": "0.1.3",
-    "solidity-coverage": "^0.6.3"
+    "@airswap/test-utils": "0.1.3"
   },
   "dependencies": {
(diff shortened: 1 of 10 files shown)


source/delegate-factory/package.json
```diff
@@ -27,9 +27,8 @@
     "verify": "truffle run verify"
   },
   "devDependencies": {
-    "@airswap/order-utils": "0.3.13",
-    "@airswap/test-utils": "0.1.3",
-    "solidity-coverage": "^0.6.3"
+    "@airswap/order-utils": "0.3.14",
+    "@airswap/test-utils": "0.1.3"
```
ejwessel: do these updates require an update of the package.json version?

source/delegate/package.json
```diff
@@ -27,9 +27,8 @@
     "verify": "truffle run verify"
   },
   "devDependencies": {
-    "@airswap/order-utils": "0.3.13",
-    "@airswap/test-utils": "0.1.3",
-    "solidity-coverage": "^0.6.3"
+    "@airswap/order-utils": "0.3.14",
+    "@airswap/test-utils": "0.1.3"
```
ejwessel: does package.json verison need to be updated?

source/delegate-factory/package.json
alicevhenshaw: Only if you change a dependency. If you change a `devDependency` then you dont need to redeploy the module. This is why `order-utils` has been bumped, as I changed the `dependency` section. The others only change `devDependency`.

PR #369, opened 2019-12-13
Move DelegateFactory to Delegate package

dmosites:


source/delegate-factory/deploys.json
```diff
@@ -1,4 +0,0 @@
-{
-  "1": "0xA55523eeD98Ba64fF7673b22963DF34E3A0305FE",
```
alicevhenshaw: do we want to keep these in the delegate? Seems like something we shouldnt lose

scripts/slither.sh
```diff
@@ -13,7 +13,7 @@
 # ------------------------------------------------------------------
 
 
- array=( indexer index delegate delegate-factory swap types wrapper )
+ array=( indexer index delegate swap types wrapper )
```
alicevhenshaw: oh. Index is still here.

dmosites: Noticed that along the way. Updated.

source/delegate-factory/deploys.json
dmosites: That's right, good call. Though this file should indicate what is actually deployed. To discuss

dmosites: Added some README content to address.

PR #375, opened 2019-12-16
add safeerc20 + prevent id/amounts validation checks in swap

poiresel:
## Description

Quick description:

- [X] Typo/small tweaks

## Changes

added SafeERC20 over using INERC20.sol
if looks good will remove INERC20.sol from tokens repo

Ensure users send only an id or amount, not both as currently supporting kind ERC20 or ERC721, not ERC1155


source/swap/test/Swap-unit.js
```diff
@@ -169,47 +169,10 @@ contract('Swap Unit Tests', async accounts => {
...
-      // auth signer to be the sender of the order
-      await swap.authorizeSender(mockSigner, {
-        from: mockSender,
-      })
-      // auth sender
-      //mock sender will take the order
-      await reverted(swap.swap(order, { from: mockSigner }), 'TRANSFER_FAILED.')
+      await reverted(
+        swap.swap(order, { from: mockSigner }),
+        'SafeERC20: ERC20 operation did not succeed.'
```
alicevhenshaw: Do you think it'll be an issue that this doesnt bubble up?

source/swap/test/Swap.js
```diff
@@ -985,7 +1009,36 @@ contract('Swap', async accounts => {
       )
     })
 
-    it('Bob buys Ticket #12345 from Alice for 1 DAI', async () => {
+    it('Bob cannot Ticket #12345 from Alice if he sends id and amount in Party struct', async () => {
```
alicevhenshaw: `Bob cannot buy Ticket #12345..`

source/swap/test/Swap-unit.js
poiresel: safeerc20 is an internal library so it will be seen, can confirm though in rinkeby for etherscan.

source/swap/test/Swap.js
```diff
@@ -985,7 +1009,36 @@ contract('Swap', async accounts => {
...
+      order.signature = await signatures.getWeb3Signature(
+        order,
+        aliceAddress,
+        swapAddress,
+        GANACHE_PROVIDER
+      )
+
+      await reverted(
+        swap(order, { from: bobAddress }),
+        'NO_AMOUNT_FIELD_IN_ERC721.'
```
alicevhenshaw: Why's there a full stop inside the revert message? Surely this test shouldnt pass with that 😢

poiresel: weird observation, if it's just a full stop, somehow truffle-assertions works, any other character, will fail it....

alicevhenshaw: That makes me sad. But good to know it still largely works.

PR #376, opened 2019-12-16
Added missing natspec line

alicevhenshaw:
## Description

Added missing natspec line, pointed out by Quantstamp.

### DIFF
--- a/source/delegate/contracts/Delegate.sol
+++ b/source/delegate/contracts/Delegate.sol
@@ -61,6 +61,7 @@ contract Delegate is IDelegate, Ownable {
     * @param delegateIndexer address Indexer contract the delegate will deploy with
     * @param delegateContractOwner address Owner of the delegate
     * @param delegateTradeWallet address Wallet the delegate will trade from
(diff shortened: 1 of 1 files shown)


source/delegate/contracts/Delegate.sol
```diff
@@ -61,6 +61,7 @@ contract Delegate is IDelegate, Ownable {
     * @param delegateIndexer address Indexer contract the delegate will deploy with
     * @param delegateContractOwner address Owner of the delegate
     * @param delegateTradeWallet address Wallet the delegate will trade from
+    * @param delegateProtocol bytes2 The protocol identifier for Delegate contracts
```
poiresel: happy to approve for natspec purposes, I think we'll need to add some more information in airswap-docs to clarify the protocol identifier meaning.

alicevhenshaw: yeah that makes sense. Was just adding this as quantstamp pointed out it was missing. We should write docs about it in more detail, I agree

PR #377, opened 2019-12-16
update packages and deploy for rinkeby

poiresel:
## Description

Quick description:

Updated packages to have a -beta.0 and deploys.json for the rinkeby deploy
Changed the ordering of yarn lint and yarn install in .circleci
Updated openzeppelin-solidity from 2.3 to 2.4

- [X] N/A

## Changes

Deploy to Rinkeby on commit hash 4677147c63109ca7202e8c03b6298cab2f8fecda
## Tests

N/A
## Test Coverage
N/A


.circleci/config.yml
```diff
@@ -24,16 +24,16 @@ jobs:
           keys:
             - v1-dependencies-{{ checksum "package.json" }}
             - v1-dependencies-
+      - run:
+          name: yarn install and dependencies
```
alicevhenshaw: nice, glad this worked 👌

source/delegate/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@airswap/delegate",
-  "version": "0.5.6",
+  "version": "0.5.6-beta.0",
```
alicevhenshaw: if you release a beta, is it normally the number before or the number after? i.e. if you have version 1 and then you will have version 2 in future, and you have a beta inbetween, should it be version 2 beta?

utils/order-utils/test/orders.js
```diff
@@ -28,7 +28,7 @@ describe('Orders', async () => {
 
   const NOV_7_2020_22_18_14 = '1604787494'
   const MAY_11_2017_00_00_00 = '1494460800'
-
+  const INVALID_KIND = '0xFFFFFF'
```
alicevhenshaw: What do you mean by 'invalid'? Just that it hasnt been registered yet?

poiresel: meaning that it is not ERC20 or ERC721, this doesn't use anything related to the transfer-registry, prior the we used for the word 'INVALID_KIND' however toBuffer requires a hex string, so I just moved it out to be a const variable.

source/delegate/package.json
poiresel: Beta in my mind wasn't related to product version. Still trying to work out how beta relates to the rest of the release process but wanted to unblock integration work.

.circleci/config.yml
ejwessel: you can probably move this back down and you won't run into the caching issue.

ejwessel: nvm.

PR #393, opened 2020-01-06
Updating revert reasons for consistency

dmosites:
Formatting remaining reasons approximately as `NOUN_ADJECTIVE`.

### DIFF
--- a/source/transfers/contracts/handlers/ERC721TransferHandler.sol
+++ b/source/transfers/contracts/handlers/ERC721TransferHandler.sol
@@ -22,7 +22,7 @@ contract ERC721TransferHandler is ITransferHandler {
     uint256 id,
     address token)
   external returns (bool) {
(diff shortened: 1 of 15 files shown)


source/transfers/contracts/handlers/ERC721TransferHandler.sol
```diff
@@ -22,7 +22,7 @@ contract ERC721TransferHandler is ITransferHandler {
     uint256 id,
     address token)
   external returns (bool) {
-    require(amount == 0, "NO_AMOUNT_FIELD_IN_ERC721");
+    require(amount == 0, "AMOUNT_MUST_BE_EMPTY_FOR_ERC721");
```
alicevhenshaw: could just be `AMOUNT_INVALID`? Like when we do `SENDER_WALLET_INVALID`? We don't need to give a complete explanation of the reason in the revert message.

PR #396, opened 2020-01-06
Remove check order from order-utils

alicevhenshaw:
## Description

Moving checkOrder to a new package to remove the cyclic dependency involving order-utils.
order-utils is no longer dependent on tokens and swap.

## Tests

Order-utils:
```
yarn run v1.19.1
$ mocha test


  Orders
    ✓ Checks that a generated order is valid

  Signatures
    ✓ Checks that a Version 0x45: personalSign signature is valid
    ✓ Checks that a Version 0x01: signTypedData signature is valid


  3 passing (27ms)
```

Debugger:
```
yarn run v1.19.1
$ mocha test --timeout 3000 --exit


  Orders
    ✓ Check correct order without signature (1053ms)
    ✓ Check correct order with signature (1017ms)
    ✓ Check expired order (951ms)
    ✓ Check invalid signature (922ms)
    ✓ Check order without allowance (850ms)
    ✓ Check NFT order without balance or allowance (1001ms)
    ✓ Check invalid token kind (708ms)
    ✓ Check NFT order without allowance (990ms)
    ✓ Check NFT order to an invalid contract (1127ms)
    ✓ Check NFT order to a valid contract (1213ms)
    ✓ Check order without balance (860ms)


  11 passing (11s)
```


utils/debugger/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@airswap/debugger",
+  "version": "0.3.17",
```
poiresel: change this to 0.0.1 as its a new package.

.circleci/config.yml
```diff
@@ -139,6 +139,17 @@ jobs:
...
+  debugger_tests:
+    <<: *container_config
+    steps:
+      - attach_workspace:
+          at: *working_directory
+      - run:
+          name: Run Tests
+          command: |
+            yarn ganache > /dev/null &
+            sleep 10s && cd utils/debugger && yarn test
```
poiresel: you can remove the sleep in order_util_tests now that it's not relying on ganache anymore.

PR #399, opened 2020-01-08
Adding signature check and more tests to order-utils

dmosites:
- `orders.isValidOrder` now includes a call to `signatures.hasValidSignature`
- Adding tests to cover more variety of invalidity

### DIFF
--- a/utils/debugger/src/check-order.js
+++ b/utils/debugger/src/check-order.js
@@ -20,8 +20,7 @@ const checkOrder = async (order, network) => {
 
   // Check the order has all necessary fields
   if (!orders.isValidOrder(order)) {
-    errors.push('Order structured incorrectly')
-    return errors
+    errors.push('Order structured incorrectly or signature invalid')
   }
 
   // Check swap address provided
@@ -69,7 +68,7 @@ const checkOrder = async (order, network) => {
 
 const checkOrderSignature = async (order, provider, errors) => {
   // Check signature is valid
-  const isValid = signatures.isSignatureValid(order)
+  const isValid = signatures.hasValidSignature(order)
   if (!isValid) {
     errors.push('Signature invalid')
   }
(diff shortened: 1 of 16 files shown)


utils/debugger/src/check-order.js
```diff
@@ -20,8 +20,7 @@ const checkOrder = async (order, network) => {
 
   // Check the order has all necessary fields
   if (!orders.isValidOrder(order)) {
-    errors.push('Order structured incorrectly')
-    return errors
+    errors.push('Order structured incorrectly or signature invalid')
```
poiresel: add push errors to line 27 as well for consistency.

utils/debugger/test/check-order.js
```diff
@@ -274,9 +255,9 @@ describe('Orders', async () => {
 
     const errors = await checker.checkOrder(order, 'rinkeby')
 
-    assert.equal(errors.length, 2)
-    assert.equal(errors[0], 'sender no NFT approval')
-    assert.equal(errors[1], 'signer is not configured to receive NFTs')
+    assert.equal(errors.length, 3)
+    assert.equal(errors[1], 'sender no NFT approval')
```
poiresel: assert the errors[0] as well

utils/order-utils/src/orders.js
```diff
@@ -21,11 +21,14 @@ const signatures = require('./signatures')
 let nonce = 100
 
 const getLatestTimestamp = async () => {
-  return (await web3.eth.getBlock('latest')).timestamp
+  if (typeof web3 !== 'undefined') {
```
poiresel: can you add a quick comment on why web3 may be undefined?

dmosites: Not all environments have a global web3.

utils/debugger/src/check-order.js
dmosites: Do you mean to remove the return?

poiresel: yes if the intended behavior is to proceed to the end and return all the errors.

PR #400, opened 2020-01-09
Add Index Verification

ejwessel:
## Description
We also need to deploy contract created contracts such that they can be verified through etherscan later.

## Changes
Added the Index to deployment

## Tests
Deployed and Verified with:
```
yarn migrate --network mainnet; yarn verify Index flatten/Flattened.sol --network mainnet
```


alicevhenshaw:
Could we not instead do the following in the migrations:
```
await deployer.deploy(Indexer, STAKING_TOKEN_ADDRESS)
await Indexer.createIndex(AST, WETH)
```

then just verify the code of the Index created by that?
Just seems a bit weird to me to create a random unusable Index for the purpose

ejwessel:
You definitely can and probably the way we should do it next time.

PR #403, opened 2020-01-13
Update swap registry

poiresel:
## Description

Add additional checks on pre-swap-checker to reflect token-registry and differentiate between erc20 and erc721. 

Quick description:
- [X] New features

## Changes

Implementation details. These are mainly to help your fellow developers.

## Tests
  26 passing (24s)

## Test Coverage
```
---------------------|----------|----------|----------|----------|----------------|
File                 |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
---------------------|----------|----------|----------|----------|----------------|
 contracts/          |      100 |      100 |      100 |      100 |                |
  Imports.sol        |      100 |      100 |      100 |      100 |                |
  PreSwapChecker.sol |      100 |      100 |      100 |      100 |                |
---------------------|----------|----------|----------|----------|----------------|
All files            |      100 |      100 |      100 |      100 |                |
---------------------|----------|----------|----------|----------|----------------|
```

### DIFF
--- a/utils/pre-swap-checker/contracts/PreSwapChecker.sol
+++ b/utils/pre-swap-checker/contracts/PreSwapChecker.sol
@@ -37,7 +37,7 @@ contract PreSwapChecker {
     bytes32 domainSeparator = Types.hashDomain(DOM_NAME, DOM_VERSION, swap);
 
     // max size of the number of errors that could exist
-    bytes32[] memory errors = new bytes32[](10);
+    bytes32[] memory errors = new bytes32[](14);
     uint8 errorCount;
 
     // Check self transfer
@@ -166,7 +166,7 @@ contract PreSwapChecker {
   /**
     * @notice Checks for valid interfaces for
     * ERC165 tokens, ERC721
-    * @param tokenAddress address address of potential ERC721 token
+    * @param tokenAddress address potential ERC721 token address
     * @return bool whether address has valid interface
     */
   function hasValidERC71Interface(
--- a/utils/pre-swap-checker/contracts/PreSwapChecker.sol
+++ b/utils/pre-swap-checker/contracts/PreSwapChecker.sol
@@ -27,8 +27,8 @@ contract PreSwapChecker {
     * @notice Takes in an order and outputs any
     * errors that Swap would revert on
     * @param order Types.Order Order to settle
-    * @param uint256 errorCount if any
-    * @param bytes32[] memory array of error messages
+    * @return uint256 errorCount if any
+    * @return bytes32[] memory array of error messages
     */
   function checkSwapSwap(
     Types.Order calldata order
--- a/utils/pre-swap-checker/contracts/PreSwapChecker.sol
+++ b/utils/pre-swap-checker/contracts/PreSwapChecker.sol
@@ -68,53 +68,54 @@ contract PreSwapChecker {
       errorCount++;
     }
 
+    // Check valid token registry handler for sender
+    if (hasValidKind(order.sender.kind, swap)) {
+      // Check the order sender
+      if (order.sender.wallet != address(0)) {
+        // The sender was specified
+        // Check if sender kind interface can correctly check balance
+        if (order.sender.kind == ERC721_INTERFACE_ID && !hasValidERC71Interface(order.sender)) {
+          errors[errorCount] = "SENDER_INVALID_ERC721";
+          errorCount++;
+        } else {
+          // Check the order sender token balance
+          if (!hasBalance(order.sender)) {
+            errors[errorCount] = "SENDER_BALANCE";
+            errorCount++;
+          }
 
-    // Check the order sender
-    if (order.sender.wallet != address(0)) {
-      // The sender was specified
+          // Check their approval
+          if (!isApproved(order.sender, swap)) {
+            errors[errorCount] = "SENDER_ALLOWANCE";
+            errorCount++;
+          }
+        }
+      }
+    } else {
+      errors[errorCount] = "SENDER_TOKEN_KIND_UNKNOWN";
+      errorCount++;
+    }
+
+     // Check valid token registry handler for signer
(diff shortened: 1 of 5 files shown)


utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -26,69 +30,125 @@ contract PreSwapChecker {
...
       }
     }
 
-    return errors;
+    return (errorCount, errors);
   }
 
+  // function to check a party has used a known kinda
+  function hasValidKind(
+    Types.Party memory party,
```
alicevhenshaw: Why not pass in just the kind instead of the whole party?

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -26,69 +30,125 @@ contract PreSwapChecker {
...
-    // Check the order signer
-    if (!hasBalance(order.signer)) {
-      errors[errorCount] = 'SIGNER_BALANCE';
+    // Check if sender kind interface can correctly check balance
+    if (order.signer.kind == ERC721_INTERFACE_ID && !hasValidERC71Interface(order.signer)) {
+      errors[errorCount] = "SIGNER_INVALID_ERC721";
       errorCount++;
+    } else {
+      if (!hasBalance(order.signer)) {
+        errors[errorCount] = "SIGNER_BALANCE";  // Check the order signe
```
alicevhenshaw: the comment doesnt make sense on this line

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -26,69 +30,125 @@ contract PreSwapChecker {
...
 
-    // Check the order signer
-    if (!hasBalance(order.signer)) {
-      errors[errorCount] = 'SIGNER_BALANCE';
+    // Check if sender kind interface can correctly check balance
+    if (order.signer.kind == ERC721_INTERFACE_ID && !hasValidERC71Interface(order.signer)) {
+      errors[errorCount] = "SIGNER_INVALID_ERC721";
       errorCount++;
+    } else {
+      if (!hasBalance(order.signer)) {
```
alicevhenshaw: We know that, because here we are in the `else` branch, we are looking at a token that is not an invalid ERC721 token - but it could be anything else. hasBalance assumes that the token is a valid ERC20 or ERC721 token - which it might not be. e.g. it could be an ERC1155 or something else. Could we be better to do a large if statement with `hasValidKind` so that checks like hasBalance and hasAllowance arent executed on tokens that then turn out to be invalid kinds anyway?

utils/pre-swap-checker/contracts/PreSwapChecker.sol
poiresel: // Check the order signer balance has been updated

utils/pre-swap-checker/contracts/PreSwapChecker.sol
poiresel: that's fair, i had made some faulty assumptions where I was thinking someone may have accidentally put the wrong kind but everything else would be ok. Having a second thought, have a nested system where we don't do balance or approvals checks until validating the kind seems preferable.

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -26,69 +30,126 @@ contract PreSwapChecker {
```
alicevhenshaw: is 10 still the correct length to be making `errors`?

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -26,69 +30,126 @@ contract PreSwapChecker {
...
-    // Check the order signer
-    if (!hasBalance(order.signer)) {
-      errors[errorCount] = 'SIGNER_BALANCE';
+    // Check valid token registry handler for sender
+    if (hasValidKind(order.sender.kind, swap)) {
+      // Check the order sender
+      if (order.sender.wallet != address(0)) {
+        // The sender was specified
+        // Check if sender kind interface can correctly check balance
+        if (order.sender.kind == ERC721_INTERFACE_ID && !hasValidERC71Interface(order.sender)) {
```
alicevhenshaw: the function `hasValidERC721Interface` already does the check that `party.kind == ERC721_INTERFACE_ID`, do that check doesnt need to be in both places

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -144,14 +163,24 @@ contract PreSwapChecker {
     return (address(tokenRegistry.transferHandlers(kind)) != address(0));
   }
 
-  // checks for valid interfaces for ERC165 tokens, ERC721
+  /**
+    * @notice Checks for valid interfaces for
+    * ERC165 tokens, ERC721
```
alicevhenshaw: I don't understand this sentence? `Checks for valid interfaces for ERC165 tokens, ERC721`

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -164,7 +193,13 @@ contract PreSwapChecker {
     return (balance >= party.amount);
   }
 
-  // function to check a party has enough allowance to swap
+  /**
+    * @notice Check a party has enough allowance to swap
+    * for ERC721 and ERC20 tokens
+    * @param party Types.Party party to check balance for
```
alicevhenshaw: allowance not balance

utils/pre-swap-checker/contracts/PreSwapChecker.sol
poiresel: the hasERC721Interfaces checks both that the token is an ERC165 and ERC721 though i'll change it as agree it's confusing. and a bad copy-paste.

PR #404, opened 2020-01-14
Wrapper swap checker

poiresel:
## Description

Quick description:

- [X] New features
- [ ] Bug fixes
- [ ] Typo/small tweaks

## Changes

Added additional checks for using the wrapper

## Tests + Coverage

  32 passing (48s)

---------------------|----------|----------|----------|----------|----------------|
File                 |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
---------------------|----------|----------|----------|----------|----------------|
 contracts/          |      100 |      100 |      100 |      100 |                |
  Imports.sol        |      100 |      100 |      100 |      100 |                |
  PreSwapChecker.sol |      100 |      100 |      100 |      100 |                |
---------------------|----------|----------|----------|----------|----------------|
All files            |      100 |      100 |      100 |      100 |                |
---------------------|----------|----------|----------|----------|----------------|

### DIFF
--- a/utils/pre-swap-checker/contracts/PreSwapChecker.sol
+++ b/utils/pre-swap-checker/contracts/PreSwapChecker.sol
@@ -48,7 +48,7 @@ contract PreSwapChecker {
   function checkSwapWrapper(
     Types.Order calldata order,
     address fromAddress,
-    address payable wrapper
+    address wrapper
     ) external view returns (uint256, bytes32[] memory ) {
     address swap = order.signature.validator;
     // max size of the number of errors that could exist
@@ -64,6 +64,7 @@ contract PreSwapChecker {
         errors[i] = swapErrors[i];
       }
     }
+
     if (order.sender.wallet != fromAddress) {
       errors[errorCount] = "MSG_SENDER_MUST_BE_ORDER_SENDER";
       errorCount++;
--- a/utils/pre-swap-checker/contracts/PreSwapChecker.sol
+++ b/utils/pre-swap-checker/contracts/PreSwapChecker.sol
@@ -27,7 +27,6 @@ contract PreSwapChecker {
 
   IWETH public wethContract;
 
-
   /**
     * @notice Contract Constructor
     * @param preSwapCheckerWethContract address
@@ -92,7 +91,6 @@ contract PreSwapChecker {
 
     // ensure that sender wallet if receiving weth has approved
     // the wrapper to transfer weth and deliver eth to the sender
-
     if (order.signer.token == address(wethContract)) {
       uint256 allowance = wethContract.allowance(order.sender.wallet, wrapper);
       if (allowance < order.signer.amount) {
@@ -100,7 +98,6 @@ contract PreSwapChecker {
         errorCount++;
       }
     }
-
     return (errorCount, errors);
   }
 
@@ -228,7 +225,6 @@ contract PreSwapChecker {
         errorCount++;
       }
     }
-
     return (errorCount, errors);
   }
 
--- a/utils/pre-swap-checker/contracts/PreSwapChecker.sol
+++ b/utils/pre-swap-checker/contracts/PreSwapChecker.sol
@@ -147,6 +147,56 @@ contract PreSwapChecker {
     return (errorCount, errors);
   }
 
+  /**
+    * @notice If order is going through wrapper to swap
+    * @param order Types.Order
+    * @param fromAddress address
+    * @param wrapper address
+    * @return uint256 errorCount if any
+    * @return bytes32[] memory array of error messages
+    */
+  function checkSwapWrapper(
+    Types.Order calldata order,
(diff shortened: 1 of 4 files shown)


utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -147,6 +149,64 @@ contract PreSwapChecker {
...
+    uint256 errorCount;
+
+    if (order.sender.wallet != fromAddress) {
+      errors[errorCount] = "MSG_SENDER_MUST_BE_ORDER_SENDER";
+      errorCount++;
+    }
+
+    // ensure that sender has approved wrapper contract on swap
+    if (!ISwap(swap).senderAuthorizations(order.sender.wallet, wrapper)) {
+      errors[errorCount] = "WRAPPER_UNAUTHORIZED";
```
alicevhenshaw: would it make more sense for the errors thrown here to match the errors the contracts would throw? SO its easy to see where they came from?

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -147,6 +149,64 @@ contract PreSwapChecker {
...
+      errorCount++;
+    }
+
+    // signature must be filled in order using Wrapper
```
alicevhenshaw: Typo in this comment

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -147,6 +149,64 @@ contract PreSwapChecker {
...
+
+    // if sender has WETH token, ensure sufficient ETH balance
+    if (order.sender.token == address(wethContract)) {
+      if (address(order.sender.wallet).balance < order.sender.amount) {
+        errors[errorCount] = "SENDER_INSUFFICIENT_ETH";
+        errorCount++;
+      }
+    }
+
+    // ensure that sender wallet if receiving weth has approve it to
```
alicevhenshaw: `approve` -> `approved`

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -147,6 +149,64 @@ contract PreSwapChecker {
...
+
+    if (order.signer.token == address(wethContract)) {
+      uint256 allowance = wethContract.allowance(order.sender.wallet, wrapper);
+      if (allowance < order.signer.amount) {
+        errors[errorCount] = "LOW_SENDER_ALLOWANCE_ON WRAPPER";
+        errorCount++;
+      }
+    }
+
+    return (errorCount, errors);
```
alicevhenshaw: It would be nice if calling `checkSwapWrapper` performed all of the necessary `swap` checks too - so you dont have to call multiple contracts. But I think that would include copying the contents of an array into another. Thoughts?

poiresel: given it's a call i may just doing the copying all the contents from one to the other. thanks!

utils/pre-swap-checker/contracts/PreSwapChecker.sol
poiresel: ok.

utils/pre-swap-checker/contracts/PreSwapChecker.sol
poiresel: Mechanism for this to work:

checkSwapSwap would be called first but with some modification. It would need access to the WETH contract address and dictate it was using the Wrapper as the balance check with the PreSwapChecker.checkSwapSwap doesn't necessarily hold for the PreSwapChecker.checkWrapperSwap. The reason is that the balance of WETH may be lower than the balance needed because the sender will be depositing ETH into the WETH contract to get a WETH balance. 

From there checkWrapperSwap can use the errorCount to determine how many slots to copy from the checkSwapErrors array into the checkWrapperSwap error array.

utils/pre-swap-checker/contracts/PreSwapChecker.sol
```diff
@@ -25,6 +25,86 @@ contract PreSwapChecker {
...
+    * @param order Types.Order
+    * @param fromAddress address
+    * @param wrapper address
+    * @return uint256 errorCount if any
+    * @return bytes32[] memory array of error messages
+    */
+  function checkSwapWrapper(
+    Types.Order calldata order,
+    address fromAddress,
+    address payable wrapper
```
alicevhenshaw: why is this payable...?

poiresel: since the wrapper contract can take in ether as a payment in the wrap function.

poiresel: was able to remove wrapper!

PR #413, opened 2020-02-10
Add `@airswap/clients` for JS Developers

dmosites:
- Adds `@airswap/clients` containing JavaScript classes for protocol components including Swap, Indexer, Server, and Delegate. TODO: Complete remaining functionality for each component.
- Adds `@airswap/constants` to consolidate network constants into a single package. TODO: Merge more constants in from other packages.
- Adds `@airswap/utils` to consolidate JavaScript utilities into a single package. TODO: Merge more utility functions in from other packages.

### DIFF
--- a/utils/clients/src/Delegate.ts
+++ b/utils/clients/src/Delegate.ts
@@ -1,0 +1,97 @@
+/*
+  Copyright 2020 Swap Holdings Ltd.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


perich:
re: TypeScript compilation tooling https://github.com/airswap/ethereum-token-metadata

dmosites:
Reviewed with team internally.

utils/clients/src/Delegate.ts
```diff
@@ -0,0 +1,97 @@
...
+      signer: {
+        token: signerToken,
+        amount: result.signerAmount.toString(),
+      },
+      sender: {
+        token: senderToken,
+        amount: result.senderAmount.toString(),
+      },
+    }
+  }
```
cloudonshore: So `getMaxQuote` on the delegate doesn't factor in the balance of the tradeWallet, so to get an accurate `maxQuote` you will need to: 
1. check if the tradeWallet has sufficient balance to cover the rule
2. If not, determine the price from the rule
3. use the tradeWallet balance * price to determine the actual `maxQuote`

Example code for doing that is here: https://github.com/airswap/AirSwap.js/blob/develop/src/delegate/index.js#L126

PR #415, opened 2020-02-10
Config updates for clients, constants, and utils packages

dmosites:


tools/utils/index.ts
```diff
@@ -79,7 +79,7 @@ export async function signOrder(
     signatory: await (await signer.getAddress()).toLowerCase(),
     validator: swapContract,
     version: signatureTypes.PERSONAL_SIGN,
-    v: v.toString(),
+    v: `${v}`,
```
perich: prefer `String(v)` for semantic clarity

tools/utils/index.ts
```diff
@@ -79,7 +79,7 @@ export async function signOrder(
     signatory: await (await signer.getAddress()).toLowerCase(),
```
perich: just `(await signer.getAddress()).toLowerCase()` should be fine

PR #420, opened 2020-02-12
Adding types

dmosites:
- Adds TS types to `@airswap/types` package
- Integrates TS types into constants, protocols, and utils
- Adds new eslint config for TS, disabled some of the new eslint rules to allow existing JS files pass until they're updated


tools/constants/index.ts
```diff
@@ -54,37 +56,40 @@ export const signatureTypes: { [key: string]: string } = {
...
   },
   AST: {
     address: '0xcc1cbd4f67cceb7c001bd4adf98451237a193ff8',
     decimals: 4,
+    symbol: 'AST',
   },
 }
 
-export const stakingTokenAddresses: { [key: string]: any } = {
+export const stakingTokenAddresses: { [key: string]: string } = {
```
syjk129: `{ [key: string]: string }` can be simplified to `Record<string, string>`

tools/ts/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type UnsignedOrder = {
+  nonce: string
+  expiry: string
+  signer: Party
+  sender: Party
+  affiliate: Party
+}
+
+export type Order = {
```
syjk129: You can make this compositional:
```
export type Order = {
  nonce: string
  expiry: string
  signer: Party
  sender: Party
  affiliate: Party
}
export type SignedOrder = Order & {
  signature: Signature
}
```

tools/ts/index.ts
```diff
@@ -0,0 +1,72 @@
+/*
```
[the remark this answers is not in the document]
dmosites: Looking into merging these into the existing `@airswap/types` package.

PR #427, opened 2020-02-13
Adding utils to get best order by lowest sender or highest signer amount

dmosites:
Get the best order by lowest sender or highest signer amount. E.g.:

```
const orders: Array<SignedOrder> = [...]
const best: SignedOrder = getByLowestSenderAmount(orders)
```


tools/utils/src/orders.ts
```diff
@@ -95,3 +96,27 @@ export async function signOrder(
     },
   }
 }
+
+export function getByLowestSenderAmount(
```
poiresel: function name change to sortBy or getOrderBy?

dmosites: Happy to change to e.g. `getOrderByLowestSenderAmount` for consistency and clarity.

tools/utils/test/orders.ts
```diff
@@ -0,0 +1,49 @@
...
+            wallet: '',
+            kind: '',
+            token: '',
+            amount: String(i + highestAmount),
+          },
+        })
+      )
+    }
+    const best = getByHighestSignerAmount(orders)
+    expect(best.sender.amount).to.equal(String(highestAmount))
```
poiresel: should this be 55? not matching on 50

tools/utils/test/orders.ts
```diff
@@ -0,0 +1,49 @@
...
+            wallet: '',
+            kind: '',
+            token: '',
+            amount: String(i + highestAmount),
+          },
+        })
+      )
+    }
+    const best = getOrderByHighestSignerAmount(orders)
+    expect(best.sender.amount).to.equal(String(highestAmount))
```
poiresel: for this test, are all the signer.Amount equivalent and just the sender Amount is changing?

tools/utils/test/orders.ts
```diff
@@ -29,21 +29,21 @@ describe('Orders', async () => {
 
   it('Checks best order by highest signer', async () => {
     const orders = []
-    const count = 5
-    const highestAmount = 50
-    for (let i = 0; i < count; i++) {
+    const highestAmount = 5
+    let count = -1
```
poiresel: this should be 0, since the ++count you want it to start at 0 not -1.

PR #436, opened 2020-02-19
Include wallet delegate quotes

dmosites:
Includes `wallet` on delegate quotes, on the sender side, using the delegate tradeWallet. This allows a taker to pull down many quotes, sort to determine the best, and, given the best quote, to provide an order to the winning delegate.

### DIFF
--- a/tools/protocols/src/Delegate.ts
+++ b/tools/protocols/src/Delegate.ts
@@ -26,8 +26,9 @@ const DelegateInterface = new ethers.utils.Interface(
 )
 
 export class Delegate {
-  address: string
   chainId: string
+  address: string
+  tradeWallet: string
   contract: ethers.Contract
 
   constructor(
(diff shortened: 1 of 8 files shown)


tools/protocols/src/Delegate.ts
```diff
@@ -124,10 +129,10 @@ export class Delegate {
   }
 
   async getQuotedOrMaxAvailable(
-    senderToken,
-    senderAmount,
-    signerToken,
-    signerAmount
+    senderToken: string,
+    senderAmount: string,
```
poiresel: on line 142, balance.lt(senderAmount) will this still work if the input is a string? same with the other math calculations that are now converted using bigNumerify.

dmosites: Yes, the argument for function `lt` is of type BigNumberish (BigNumber | string | number | Arrayish)

tools/protocols/src/Delegate.ts
```diff
@@ -26,17 +26,19 @@ const DelegateInterface = new ethers.utils.Interface(
 )
 
 export class Delegate {
-  address: string
   chainId: string
+  address: string
```
syjk129: Ideally, these should have accessibility modifiers

PR #446, opened 2020-02-20
Add metadata

perich:
Adds a TokenMetadata class that loads metadata from MetaMask and Idex on instantiation. Token data can also be fetched directly from the deployed contract with `crawlToken()`


.eslintrc.js
```diff
@@ -35,5 +35,6 @@ module.exports = {
     '@typescript-eslint/camelcase': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
     '@typescript-eslint/no-var-requires': 0,
+    '@typescript-eslint/explicit-member-accessibility': 0,
```
dmosites: This negates the first rule above (where it's set to 1) so if we'd like to change let's discuss

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
```
dmosites: Should likely add `.vscode` to `.gitignore`

perich: If we commit it, then everyone who is using .vscode with this repo will have their editor configured to use the local TypeScript installation for syntax highlighting, vs whichever one they have built in to their version of VSCode

dmosites: Okay, that's a benefit. To discuss with the team

tools/metadata/lib/index.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // get token objects in an array
+  public getTokens = (): NormalizedToken[] => this.tokens
+
+  // get token objects in an object keyed by address
+  public getTokensByAddress = (): { [address: string]: NormalizedToken } =>
+    this.tokensByAddress
+
+  private _storeTokensByAddress = () => {
```
syjk129: What do you think about adding member ordering lint rules? I was thinking of adding this to trader and explorer soon. https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md

tools/metadata/lib/index.ts
```diff
@@ -0,0 +1,173 @@
...
+    const { tokensByAddress, tokens } = this
+
+    tokens.forEach(token => {
+      if (
+        !Object.prototype.hasOwnProperty.call(tokensByAddress, token.symbol)
```
syjk129: Shouldn't this be `token.address`?

Also, what's your reasoning for using hasOwnProperty vs checking if the index exists? imo this is easier to read
```
if (!tokensByAddress[token.address]) {
  tokensByAddress[token.address] = token
}
```

tools/metadata/lib/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  // get token objects in an array
+  public getTokens = (): NormalizedToken[] => this.tokens
+
+  // get token objects in an object keyed by address
+  public getTokensByAddress = (): { [address: string]: NormalizedToken } =>
+    this.tokensByAddress
+
+  private _storeTokensByAddress = () => {
+    const { tokensByAddress, tokens } = this
```
syjk129: Can we be consistent whether we deconstruct `this` or not? It looks like you use both styles here. Just a personal preference but I like to not deconstruct objects where it makes sense to know where the variable comes from.

tools/metadata/lib/helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+        .then((bytes32: Arrayish) => ethers.utils.parseBytes32String(bytes32))
+    )
+}
+
+export async function getTokenDecimals(tokenAddress: string) {
+  if (!tokenAddress || !tokenAddress.startsWith('0x')) {
+    throw Error(`Invalid 'tokenAddress' parameter`)
+  }
+
+  return getContract(tokenAddress, ERC20_ABI).decimals()
```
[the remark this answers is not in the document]
perich: it's not necessary here, the BYTES_32 fallback is only for symbol and name

tools/metadata/lib/index.ts
perich: good idea cc @dmosites

tools/metadata/lib/index.ts
perich: Usually I will just deconstruct it if there is more than 1 variable, or I'm going to reference that variable a lot. Typing `this` over and over just creates more surface area for bugs imo. I don't feel that strongly about it though, so I'm find doing whatever

tools/metadata/lib/helpers.ts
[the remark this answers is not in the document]
syjk129: ah gotcha

tools/metadata/lib/index.ts
perich: It's an `airbnb-config-eslint` lint rule that I've just picked up a habit.


Consider this example
```
Object.prototype.baz = 100;
var foo = {"bar": 1}


// will be false because in foo object there is no baz property 
// hasOwnProperty checks that the object has the specified property 
// and does not check that property in the prototype chain
if (foo.hasOwnProperty('baz')) {
  console.log('foo.hasOwnProperty("baz")');
}


//  will be true because baz property will be found 
//  in parent object through prototype chain
if (foo.baz) {
  console.log('foo.baz');
}

```

tools/metadata/lib/index.ts
dmosites: Sure, sounds good 👍

perich: added

tools/metadata/lib/index.ts
syjk129: Yeah I don't really mind too but in cases where I have a local variable that has the same meaning/context as one from props or an object, I like to see the clear distinction. We can ponder on this though

PR #469, opened 2020-02-27
use checksummed address when looking up trust img

perich:


tools/metadata/lib/index.ts
```diff
@@ -60,7 +61,9 @@ class TokenMetadata {
         address: token.address,
         decimals: token.decimals,
         symbol: token.symbol,
-        image: `${TRUST_WALLET_IMAGE_API}/${token.address}/logo.png`,
+        image: `${TRUST_WALLET_IMAGE_API}/${ethers.utils.getAddress(
```
dmosites: Best to consolidate `this.getImageURL(token.address)`

tools/metadata/lib/index.ts
```diff
@@ -139,12 +142,14 @@ class TokenMetadata {
...
-    return `${TRUST_WALLET_IMAGE_API}/${address}/logo.png`
+    return `${TRUST_WALLET_IMAGE_API}/${ethers.utils.getAddress(
+      address
+    )}/logo.png`
   }
 
   // this will fail if the token you search isn't present in the Trust Wallet metadata, or if the letter casing doesn't match Trust's metadata
   public fetchImageBinaryUnstable = (address: string): Promise<string> => {
-    return axios.get(this.getImageURL(address))
+    return axios.get(this.getImageURL(ethers.utils.getAddress(address)))
```
dmosites: `getAddress` will have been called twice

PR #472, opened 2020-03-03
Adds validation and tests for Server

dmosites:
* Adds validation to Server client
* Adds Server client tests
* `TODO`: Add tests for remaining clients


tools/utils/src/orders.ts
```diff
@@ -73,25 +63,91 @@ export function createOrderForQuote(
...
+      signature['r'],
+      signature['s']
+    )
+  } catch (e) {
+    return false
+  }
+  const signingAddress = ethUtil.bufferToHex(
+    ethUtil.pubToAddress(signingPubKey)
+  )
+  return signingAddress.toLowerCase() == signature['signatory']
```
perich: non strict equality check

PR #474, opened 2020-03-10
Update Swap tests to use @airswap/utils

dmosites:


tools/utils/index.ts
```diff
@@ -118,9 +118,10 @@ export function lowerCaseAddresses(obj: any): any {
   for (const key in obj) {
     if (typeof obj[key] === 'object') {
       lowerCaseAddresses(obj[key])
-    }
-    if (typeof obj[key] === 'string' && obj[key].indexOf('0x') === 0) {
+    } else if (typeof obj[key] === 'string' && obj[key].indexOf('0x') === 0) {
```
poiresel: what's the purpose for the else addition?

dmosites: Only checks the second condition if the first fails, better logic.

PR #477, opened 2020-03-11
Replaces @airswap/order-utils with @airswap/utils in most cases

dmosites:


source/transfers/package.json
```diff
@@ -17,7 +17,6 @@
     "test": "truffle test"
   },
   "devDependencies": {
-    "@airswap/order-utils": "0.3.20",
     "@airswap/test-utils": "0.1.6",
```
poiresel: do you need to add @airswap/constants here?

source/delegate/package.json
```diff
@@ -26,15 +26,16 @@
...
-    "@airswap/transfers": "1.1.3",
-    "solidity-coverage": "^0.6.3"
-  },
   "dependencies": {
     "@airswap/indexer": "3.6.9",
     "@airswap/swap": "5.4.8",
     "openzeppelin-solidity": "2.4"
+  },
+  "devDependencies": {
+    "@airswap/constants": "0.3.5",
```
poiresel: @airswap/utils needs to be added

dmosites:
> 2 missing dependencies from what I can tell. otherwise looks good!

Added!

PR #483, opened 2020-03-12
Automatically write to deploys.json during deployment

ejwessel:
## Description

Automatically write to deploys.json during deployment
Note: this doesn't handle updating package versions

## Changes
Upon deployment it will find the correct deploys.json for the respective module and update it with the new address


tools/deployer/migrations/1_initial_migration.js
```diff
@@ -67,14 +89,22 @@ module.exports = async (deployer, network) => {
...
-  await deployer.deploy(
-    Delegate,
-    Swap.address,
-    Indexer.address,
-    ADDRESS_ZERO,
-    ADDRESS_ZERO,
-    '0x0001'
-  )
+  //Deploy Factory-Created Contracts that also need to be verified
+  if (network !== "DEVELOPMENT") {
```
dmosites: Could merge this and the following conditional. Under what circumstances will network === 'DEVELOPMENT'?

ejwessel: > Under what circumstances will network === 'DEVELOPMENT'?
Localhost

PR #488, opened 2020-03-12
Final code style pass

dmosites:
* Runs all sol, ts, js files through new prettier config
* Adds prettier check to pre-commit and circleci


source/swap/.solcover.js
```diff
@@ -1,13 +1,13 @@
 module.exports = {
-    port: 8545,
```
poiresel: remove this file from the commit.

poiresel: can remove this file as coverage doesn't need it.

PR #503, opened 2020-03-31
add support for ERC721 tokens in metadata package

perich:
This PR uses the OpenSea API to fetch ERC721 tokens and store them in the memory tokens store.

### DIFF
--- a/tools/metadata/src/constants.ts
+++ b/tools/metadata/src/constants.ts
@@ -7,6 +7,8 @@ export const TRUST_WALLET_IMAGE_API =
 export const METAMASK_IMAGE_API =
   'https://raw.githubusercontent.com/MetaMask/eth-contract-metadata/master/images'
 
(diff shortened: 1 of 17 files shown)


tools/metadata/src/constants.ts
```diff
@@ -251,18 +251,21 @@ export const rinkebyTokensByAddress: Record<string, NormalizedToken> = {
     address: '0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea',
     decimals: 18,
     symbol: 'DAI',
+    kind: 'ERC20',
```
dmosites, line 254: Would be `0x36372b07 ` but can just use `tokenKinds.ERC20` from constants

perich: This is meant to be a string representation either `'ERC20'` | `'ERC721'` to better match the schema used in the legacy hosted metadata service. It's used by the UI and by AirSwap.js to determine how a piece of token metadata should be treated

PR #504, opened 2020-03-31
Delegate v2

alicevhenshaw:
## Description

- Multiple rules for the same pair on the same delegate
- getMaxQuote takes into account tradeWallet balance and allowance
- Rules are referenced by ID
- Rules are priced using senderAmount and signerAmount

## Tests:
`  184 passing (1m)`

## Coverage
Screenshot 2020-04-13 at 3 09 17 PM


source/delegate/contracts/interfaces/IDelegateV2.sol
```diff
@@ -0,0 +1,102 @@
...
+    uint256 senderAmount,
+    uint256 signerAmount
+  );
+
+  event DeleteRule(
+    address indexed owner,
+    uint256 indexed ruleID
+  );
+
+  event ProvideOrder(
```
cloudonshore, line 48: So the important thing to remember when it comes to rule creation is that there's going to be a lot of need for lifecycle tracking. Lifecycle events include
- when a rule is created
- when a rule is deleted
- when a rule is filled against by an order

I think because of this, in cases where an order is filling against multiple rules, a ProvideOrder event will need to be emitted for each rule that an order fills against, with the senderAmount/signerAmount that each rule filled against the order. The ruleID will also need to be included in each provideOrder event. 

The provideOrder event is information for the delegate owner, not for the sender of the transaction. The delegate owner cares about the status of their individual rules, not the aggregate volume of a transaction against them.

poiresel:
Two quick suggestions:
* Append RuleID to the events
* Append both tradeWallet + owner to events

source/delegate/contracts/interfaces/IDelegateV2.sol
alicevhenshaw: Yep that makes sense.

Do you think it would be better to have 2 events:
- RuleFilled(ruleID, senderAmount, signerAmount, etc)
- ProvideOrder(totalSenderAmount, totalSignerAmount, etc)

Is the combined ProvideOrder information in the above example needed? Or is just the RuleFilled information enough?

cloudonshore: I think all I need is the RuleFilled events.

poiresel: In addition, outside of just an event, is it worth storing information in a mapping to keep filled orders on the delegate?

poiresel:
Separation of the Double Linked List logic from the pricing logic would help with code readability and modularity

source/delegate/contracts/DelegateV2.sol
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  function deleteRule(uint256 ruleID) external {
+    _deleteRule(ruleID);
+  }
+
+  function provideOrder(Types.Order calldata order) external {
+    uint256 ruleID = firstRuleID[order.sender.token][order.signer.token];
+
+    require(order.signature.v != 0, "SIGNATURE_MUST_BE_SENT");
```
dmosites, line 150: Do we still need to require signature here?

alicevhenshaw: The security review of that was another point on the list of things to consider for V2, but I haven't done that analysis so didn't want to remove it. Happy to remove it once we've sat down and gone over all of the consequences of removing it

source/delegate/contracts/DelegateV2.sol
```diff
@@ -0,0 +1,543 @@
...
+    require(
+      order.sender.kind == ERC20_INTERFACE_ID,
+      "SENDER_KIND_MUST_BE_ERC20"
+    );
+
+    // Ensure that a rule exists.
+    require(ruleID != NO_RULE, "TOKEN_PAIR_INACTIVE");
+
+    uint256 signerAmount = order.signer.amount;
+    uint256 senderAmount = order.sender.amount;
```
dmosites, line 186: Suggest `remainingSignerAmount` and `remainingSenderAmount` as below

source/delegate/contracts/DelegateV2.sol
```diff
@@ -0,0 +1,543 @@
...
+    _setIntent(senderToken, signerToken, newStakeAmount);
+  }
+
+  function setIntent(
+    address senderToken,
+    address signerToken,
+    uint256 newStakeAmount
+  ) external onlyOwner {
+    // check whether the delegate has a rule for that market
+    require(firstRuleID[senderToken][signerToken] != NO_RULE, "NO_RULE_EXISTS");
```
dmosites, line 140: Suggest `RULE_MUST_EXIST`

source/delegate/contracts/DelegateV2.sol
```diff
@@ -0,0 +1,543 @@
...
+          rules[ruleID].signerAmount
+        );
+
+        ruleID = rules[ruleID].nextRuleID;
+        _deleteRule(rules[ruleID].prevRuleID);
+      } else {
+        // only a fraction of this rule is needed so we calculate the signer amount
+        // neither divisions can have a denominator of 0. removing safemath preserves some calculation accuracy
+        uint256 numerator = rules[ruleID].signerAmount.mul(senderAmount);
+        uint256 signerFraction = numerator.div(rules[ruleID].senderAmount);
```
dmosites, line 208: Suggest `signerAmountFraction`

source/delegate/contracts/DelegateV2.sol
```diff
@@ -0,0 +1,543 @@
...
+      "SENDER_KIND_MUST_BE_ERC20"
+    );
+
+    // Ensure that a rule exists.
+    require(ruleID != NO_RULE, "TOKEN_PAIR_INACTIVE");
+
+    uint256 signerAmount = order.signer.amount;
+    uint256 senderAmount = order.sender.amount;
+
+    while (ruleID != NO_RULE && senderAmount > 0) {
```
dmosites, line 188: Suggest consistent logic (rule check first) across the three similar `while` loops here and below

source/delegate/contracts/DelegateV2.sol
```diff
@@ -0,0 +1,543 @@
...
+    uint256 remainingSenderAmount;
+    (signerAmount, remainingSenderAmount) = _getSignerSideQuote(
+      senderBalance,
+      senderToken,
+      signerToken
+    );
+    senderAmount = senderBalance.sub(remainingSenderAmount);
+  }
+
+  function getLevels(address senderToken, address signerToken)
```
dmosites, line 310: Suggest `getRules` here

alicevhenshaw: Done

source/delegate/contracts/DelegateV2.sol
alicevhenshaw: Done

source/delegate/contracts/DelegateV2.sol
alicevhenshaw: Done

source/delegate/contracts/DelegateV2.sol
alicevhenshaw: Done

source/delegate/contracts/DelegateV2.sol
alicevhenshaw: Done

source/delegate/test/Delegate-unit.js
```diff
@@ -514,24 +514,6 @@ contract('Delegate Unit Tests', async accounts => {
...
-      )
-
-      //mock a successful transfer
-      await mockStakingToken.givenMethodReturnBool(
-        mockStakingToken_transfer,
-        true
-      )
-
-      await passes(delegate.unsetRuleAndIntent(MOCK_WETH, MOCK_DAI))
-    })
```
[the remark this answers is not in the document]
alicevhenshaw: Its an exact duplicate of the test above it

source/delegate/test/Delegate-V2-unit.js
```diff
@@ -0,0 +1,1372 @@
...
+  afterEach(async () => {
+    await revertToSnapshot(snapshotId)
+  })
+
+  before('Setup DelegateV2 Contract', async () => {
+    await setupMockSwap()
+    await setupMockTokens()
+    await setupMockIndexer()
+
+    await mockStakingToken.givenMethodReturnBool(mockToken_approve, true)
```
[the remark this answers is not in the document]
alicevhenshaw: yes good point

source/delegate/test/Delegate-V2-unit.js
```diff
@@ -0,0 +1,1372 @@
...
+        mockIndexer.address,
+        notOwner,
+        tradeWallet,
+        PROTOCOL,
+        {
+          from: owner,
+        }
+      )
+
+      // being provided an empty address, it should leave the owner unchanged
```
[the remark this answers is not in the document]
alicevhenshaw: no - removed it

ejwessel:
Weird, my other comment never appeared. What's the plan with integrating this with the factory?

PR #512, opened 2020-04-15
Natspec delegate v2

alicevhenshaw:
## Description

Quick description:

- [X] N/A
- [ ] New features
- [ ] Bug fixes
- [ ] Typo/small tweaks

### DIFF
--- a/source/delegate/contracts/DelegateV2.sol
+++ b/source/delegate/contracts/DelegateV2.sol
@@ -107,7 +107,7 @@ contract DelegateV2 is IDelegateV2, Ownable {
    * @notice Creates a new trading rule
    * @dev The price is determined by the ratio of senderAmount to signerAmount
    * @param senderToken address The token the delegate owner will supply in an order
-   * @param signerToken address The token they want to trade with
+   * @param signerToken address The token the user will supply
    * @param senderAmount uint256 The total number of sender tokens they want to receive
    * @param signerAmount uint256 The total number of signer tokens they want to send
    */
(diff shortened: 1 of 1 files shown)


source/delegate/contracts/DelegateV2.sol
```diff
@@ -145,6 +175,10 @@ contract DelegateV2 is IDelegateV2, Ownable {
     _setIntent(senderToken, signerToken, newStakeAmount);
   }
 
+  /**
+   * @notice Deletes a rule and unsets intent to trade on that market on the indexer
```
dmosites, line 179: * trade that pair on the indexer

source/delegate/contracts/DelegateV2.sol
```diff
@@ -269,6 +320,13 @@ contract DelegateV2 is IDelegateV2, Ownable {
     if (remainingSenderAmount > 0) return 0;
   }
 
+  /**
+   * @notice Get the signer-side quote for a trade
```
dmosites, line 324: * sender-side

source/delegate/contracts/DelegateV2.sol
```diff
@@ -269,6 +320,13 @@ contract DelegateV2 is IDelegateV2, Ownable {
...
+   * @param signerAmount uint256 The amount of tokens they want to send
```
dmosites, line 325: Might as well use "supply" for each side, to avoid confusion with "sender" language.
For example, here would be "...amount of tokens the user will supply"

alicevhenshaw: Yep I like that

PR #519, opened 2020-04-28
Add Block Number and Update Entity Namings

ejwessel:
## Description
Adding block numbers to `ExecutedOrders` as it will allow us to properly alert of trades from the Block Cruncher Service
https://github.com/airswap/block-data-cruncher/pull/26


ejwessel:
* mainnet
* rinkeby
* goerli
* kovan
have all been deployed to.

PR #520, opened 2020-04-28
Upgrades solidity-coverage with some test modifications

dmosites:
- `yarn coverage` can now be run without having to modify source code
- Tests explicitly set expiry based on block timestamp
- Providers connect to the injected web3 provider host


tools/deployer/truffle-config.js
```diff
@@ -67,7 +67,7 @@ module.exports = {
       },
     },
   },
-  plugins: ['truffle-verify', 'truffle-flatten'],
+  plugins: ['truffle-verify', 'truffle-flatten', 'solidity-coverage'],
```
ejwessel, line 70: don't need `solidity-coverage` in this file

PR #535, opened 2020-05-15
Function to find token by symbol for @airswap/metadata

dmosites:
Adds a function to return all tokens sharing a symbol. For example:

```
await new TokenMetadata().fetchKnownTokens()
const DAI = metadata.findTokensBySymbol('DAI').shift()
```

### DIFF
--- a/tools/metadata/index.ts
+++ b/tools/metadata/index.ts
@@ -170,7 +170,7 @@ class TokenMetadata {
   }
 
   // get token objects with symbols that match a query
-  public findBySymbol(query: string): NormalizedToken[] {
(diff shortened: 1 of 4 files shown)


tools/metadata/index.ts
```diff
@@ -169,6 +169,17 @@ class TokenMetadata {
     return this.tokensByAddress
   }
 
+  // get token objects with symbols that match a query
+  public findBySymbol(query: string): NormalizedToken[] {
```
cloudonshore, line 173: Might be more semantically clear to call it `findTokensBySymbol`

PR #545, opened 2020-10-21
Updating IDEX asset URL and response handler for @airswap/metadata

dmosites:
Changes to the `@airswap/metadata` package:

* IDEX asset URL has been updated
* IDEX response data format has been updated
* IDEX response is now an array

New URL: https://api.idex.io/v1/assets


cloudonshore (approved):
lgtm

PR #546, opened 2020-12-02
Adds sub-folders to source for organization

dmosites:
Structures the contents of `source/` into:
```
source/
    core/
        swap/
        transfers/
        types/
    delegation/
        delegate/
    discovery/
        indexer/
    helpers/
        balances/
        tokens/
        validator/
        wrapper/
````


dmosites:
Decided against this and to stick to a flat structure and revisit as needed.

PR #549, opened 2020-12-14
Adding Locker and Pool contracts (AST+)

dmosites:


source/locker/contracts/Locker.sol
```diff
@@ -0,0 +1,180 @@
...
+    emit SetThrottlingPercentage(_throttlingPercentage);
+    emit SetThrottlingDuration(_throttlingDuration);
+    emit SetThrottlingBalance(_throttlingBalance);
+  }
+
+  /**
+   * @notice Locks tokens from the msg.sender.
+   * @param amount Amount of tokens to lock
+   */
+  function lock(uint256 amount) public {
```
vdrg, line 103: maybe the logic for both lock and lockFor could be encapsulated in an internal _lock(address, uint256), just to deduplicate some code

source/pool/contracts/Pool.sol
```diff
@@ -0,0 +1,161 @@
...
+import "@openzeppelin/contracts/cryptography/MerkleProof.sol";
+
+/**
+ * @title Pool: Claim Tokens Based on a Pricing Function
+ */
+contract Pool is Ownable, Pausable {
+  using SafeMath for uint256;
+
+  // Higher the scale, lower the output for a claim
+  uint256 public immutable scale;
```
vdrg, line 33: I would use uppercase for SCALE and MAX so that it is easier to understand that they are constants

source/pool/contracts/Pool.sol
```diff
@@ -0,0 +1,161 @@
...
+    IERC20(token).transfer(msg.sender, amount);
+    emit Single(root, msg.sender, token, amount);
+  }
+
+  /**
+   * @notice Bulk claim of tokens to be transferred to msg.sender
+   * @param claims Claim[]
+   * @param token address
+   */
+  function bulk(Claim[] memory claims, address token) public {
```
vdrg, line 108: It feels a little bit inconsistent to use Claim here but using the Claim fields directly in `single`. One option would be to also use the Claim struct in `single`, or to just remove `single` and rename `bulk` to `claim` and use that for everything

source/pool/contracts/Pool.sol
```diff
@@ -0,0 +1,161 @@
...
+   * @param proof bytes32[]
+   */
+  function verify(
+    address participant,
+    bytes32 root,
+    uint256 score,
+    bytes32[] memory proof
+  ) public view returns (bool valid) {
+    bytes32 leaf = keccak256(abi.encodePacked(participant, score));
+    return MerkleProof.verify(proof, root, leaf);
```
vdrg, line 159: this is 100% optional, but doing `using MerkleProof for bytes32[]`, you could then use `proof.verify(root, leaf)`. Maybe it is more explicit to use `MerkleProof.verify` though, so not sure if the change is worth it.

vdrg (commented):
Looks good, just noted a few optional small changes

source/pool/contracts/Pool.sol
dmosites: We might want to make these mutable so admins can adjust if conditions change. Will take a look.

source/pool/contracts/Pool.sol
```diff
@@ -0,0 +1,161 @@
...
+   * @param participant address
+   * @param root bytes32
+   * @param score uint256
+   * @param proof bytes32[]
+   */
+  function verify(
+    address participant,
+    bytes32 root,
+    uint256 score,
+    bytes32[] memory proof
```
ejwessel, line 156: I'm sure OZ thought of this, but one concern I have (despite the complexity growing by log(n)) is that the proof is too large to be computed over. I don't see is hitting this limit any time soon.

vdrg: maybe if the merkle tree ends up being too big, it can be divided into 2 or more trees and provide multiple roots

ejwessel: Yeah, actually that's an interesting point. @dmosites is each vote it's own merkle?

dmosites: Figure with 10K leaves (that's a lot of voters) the proof lands around 10 hashes.

source/pool/contracts/Pool.sol
dmosites: Agree it's more explicit, will stick with `MerkleProof.verify`.

source/pool/contracts/Pool.sol
```diff
@@ -0,0 +1,156 @@
...
+    bytes32[] proof;
+  }
+
+  /**
+   * @notice Constructor
+   * @param _scale uint256
+   * @param _max uint256
+   */
+  constructor(uint256 _scale, uint256 _max) public {
+    require(_max <= 100, "MAX_TOO_HIGH");
```
ejwessel, line 72: nit: 100 as const?

source/pool/contracts/Pool.sol
```diff
@@ -0,0 +1,156 @@
...
+    return MerkleProof.verify(proof, root, leaf);
+  }
+
+  function setScale(uint256 _scale) public onlyOwner {
+    scale = _scale;
+    emit SetScale(scale);
+  }
+
+  function setMax(uint256 _max) public onlyOwner {
+    require(_max <= 100, "MAX_TOO_HIGH");
```
ejwessel, line 152: nix: 100 as const?

ejwessel (approved):
LGTM. just nits. Will leave to you to in order to decide they're worthy changes.

PR #550, opened 2020-12-18
Metadata providers

dmosites:
Updates `@airswap/metadata` to take an ethers provider instead of a chain ID.

This makes custom providers possible. Version updated from `0.1.11` to `0.2.0`.

### DIFF
--- a/tools/metadata/src/helpers.ts
+++ b/tools/metadata/src/helpers.ts
@@ -1,6 +1,5 @@
 import * as ethers from 'ethers'
 import { ERC20_ABI, ERC20_BYTES32_ABI } from './constants'
-import { Arrayish } from 'ethers/utils'
 
 export function getContract(
   address: string,
--- a/tools/metadata/src/helpers.ts
+++ b/tools/metadata/src/helpers.ts
@@ -2,50 +2,58 @@ import * as ethers from 'ethers'
 import { ERC20_ABI, ERC20_BYTES32_ABI } from './constants'
 import { Arrayish } from 'ethers/utils'
 
-export function getContract(address: string, ABI: any, network = 'mainnet') {
-  const provider =
-    network === 'mainnet'
-      ? ethers.getDefaultProvider()
-      : ethers.getDefaultProvider(network.toLowerCase())
+export function getContract(
+  address: string,
+  ABI: any,
+  provider: ethers.providers.BaseProvider
+) {
   if (!address || !address.startsWith('0x')) {
     throw Error(`Invalid 'address' parameter '${address}'.`)
   }
-
   return new ethers.Contract(address, ABI, provider)
 }
 
-export async function getTokenName(tokenAddress: string, network: string) {
+export async function getTokenName(
+  tokenAddress: string,
+  provider: ethers.providers.BaseProvider
+) {
   if (!tokenAddress || !tokenAddress.startsWith('0x')) {
     throw Error(`Invalid 'tokenAddress' parameter`)
   }
 
-  return getContract(tokenAddress, ERC20_ABI, network)
+  return getContract(tokenAddress, ERC20_ABI, provider)
     .name()
     .catch(() =>
-      getContract(tokenAddress, ERC20_BYTES32_ABI, network)
+      getContract(tokenAddress, ERC20_BYTES32_ABI, provider)
         .name()
         .then((bytes32: Arrayish) => ethers.utils.parseBytes32String(bytes32))
     )
 }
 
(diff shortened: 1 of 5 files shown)


tools/metadata/src/helpers.ts
```diff
@@ -2,50 +2,58 @@ import * as ethers from 'ethers'
...
+export async function getTokenName(
+  tokenAddress: string,
+  provider: ethers.providers.BaseProvider
+) {
   if (!tokenAddress || !tokenAddress.startsWith('0x')) {
     throw Error(`Invalid 'tokenAddress' parameter`)
   }
 
-  return getContract(tokenAddress, ERC20_ABI, network)
+  return getContract(tokenAddress, ERC20_ABI, provider)
```
vdrg, line 24: I would replace this with `await getContract(...).name()`, wrapped in a try catch

tools/metadata/src/helpers.ts
```diff
@@ -2,50 +2,58 @@ import * as ethers from 'ethers'
...
+export async function getTokenSymbol(
+  tokenAddress: string,
+  provider: ethers.providers.BaseProvider
+) {
   if (!tokenAddress || !tokenAddress.startsWith('0x')) {
     throw Error(`Invalid 'tokenAddress' parameter`)
   }
 
-  return getContract(tokenAddress, ERC20_ABI, network)
+  return getContract(tokenAddress, ERC20_ABI, provider)
```
vdrg, line 41: same as above comment, currently this function doesn't need to be async as you are returning a promise, but to keep it consistent with the rest of the code I would keep the async and use await

tools/metadata/src/helpers.ts
```diff
@@ -2,50 +2,58 @@ import * as ethers from 'ethers'
...
     .catch(() =>
-      getContract(tokenAddress, ERC20_BYTES32_ABI, network)
+      getContract(tokenAddress, ERC20_BYTES32_ABI, provider)
         .symbol()
         .then((bytes32: Arrayish) => ethers.utils.parseBytes32String(bytes32))
     )
 }
 
-export async function getTokenDecimals(tokenAddress: string, network: string) {
+export async function getTokenDecimals(
```
vdrg, line 50: this function is fine but doesn't need to be async as you are already returning a promise

tools/metadata/src/helpers.ts
```diff
@@ -20,14 +19,13 @@ export async function getTokenName(
...
-  return getContract(tokenAddress, ERC20_ABI, provider)
-    .name()
-    .catch(() =>
-      getContract(tokenAddress, ERC20_BYTES32_ABI, provider)
-        .name()
-        .then((bytes32: Arrayish) => ethers.utils.parseBytes32String(bytes32))
+  try {
+    return await getContract(tokenAddress, ERC20_ABI, provider).name()
+  } catch {
+    return ethers.utils.parseBytes32String(
```
vdrg, line 25: again just being picky 😅  but maybe it would be a little bit more readable if you do:

```
const name = await getContract(tokenAddress, ERC20_BYTES32_ABI, provider).name()
return ethers.utils.parseBytes32String(name)
```

tools/metadata/src/helpers.ts
```diff
@@ -38,13 +36,13 @@ export async function getTokenSymbol(
...
-    .symbol()
-    .catch(() =>
-      getContract(tokenAddress, ERC20_BYTES32_ABI, provider)
-        .symbol()
-        .then((bytes32: Arrayish) => ethers.utils.parseBytes32String(bytes32))
+  try {
+    return await getContract(tokenAddress, ERC20_ABI, provider).symbol()
+  } catch {
+    return ethers.utils.parseBytes32String(
+      await getContract(tokenAddress, ERC20_BYTES32_ABI, provider).symbol()
```
vdrg, line 43: same as above

tools/metadata/src/helpers.ts
```diff
@@ -55,5 +53,5 @@ export async function getTokenDecimals(
     throw Error(`Invalid 'tokenAddress' parameter`)
   }
 
-  return getContract(tokenAddress, ERC20_ABI, provider).decimals()
+  return await getContract(tokenAddress, ERC20_ABI, provider).decimals()
```
vdrg, line 56: I explained incorrectly my bad, I meant that you can make this function not async, just `export function getTokenDecimals(...)`, and do the normal return as before `return getContract(tokenAddress, ERC20_ABI, provider).decimals()`

vdrg (approved):

PR #552, opened 2020-12-29
Adds light order support to @airswap/types and @airswap/utils

dmosites:


tools/utils/src/orders.ts
```diff
@@ -244,3 +247,46 @@ export function isValidOrder(order: Order): boolean {
   }
   return false
 }
+
+export function createLightOrder({
+  nonce = Date.now().toString(),
+  expiry = Math.round(Date.now() / 1000 + SECONDS_IN_DAY).toString(),
+  signerToken = ADDRESS_ZERO,
```
vdrg, line 254: Maybe we should not set default values for tokens, wallets and amounts, as the user should always provide these

ejwessel (approved):
besides vicente's comment this LGTM.

tools/utils/src/orders.ts
dmosites: The purpose of this function is to create a complete order from partial values, so think we should keep it as-is, but could rename to `createLightOrderWithDefaults`. Thoughts?

dmosites: Going to remove this and use an object literal in the test.

PR #575, opened 2021-02-10
feat: support contract wallets, split signatures and add fees

vdrg:
TODO: Add additional checks and tests

## Description

Quick description:

- [ ] N/A
- [X] New features
- [ ] Bug fixes
- [ ] Typo/small tweaks


source/light/contracts/Light.sol
```diff
@@ -93,51 +104,65 @@ contract Light is ILight {
...
+    senderToken.safeTransferFrom(msg.sender, signerWallet, senderAmount);
 
     // Transfer token from signer to sender.
-    signerToken.safeTransferFrom(signer, msg.sender, signerAmount);
+    signerToken.safeTransferFrom(signerWallet, msg.sender, signerAmount);
+
+    // Transfer fee
+    if (fee > 0) {
+      uint256 feeAmount = signerAmount.mul(fee).div(FEE_DIVISOR);
+      signerToken.safeTransferFrom(signerWallet, feeWallet, feeAmount);
```
wjmelements, line 159: Also check if `feeAmount` is greater than 0. For gas tokens might be zero more often.

source/light/contracts/Light.sol
```diff
@@ -93,51 +104,65 @@ contract Light is ILight {
...
+      require(authorized[signerWallet] == signer, "UNAUTHORIZED");
+    }
+
     // Transfer token from sender to signer.
-    senderToken.safeTransferFrom(msg.sender, signer, senderAmount);
+    senderToken.safeTransferFrom(msg.sender, signerWallet, senderAmount);
 
     // Transfer token from signer to sender.
-    signerToken.safeTransferFrom(signer, msg.sender, signerAmount);
+    signerToken.safeTransferFrom(signerWallet, msg.sender, signerAmount);
```
wjmelements, line 154: You should add another method that allows sender to configure the destination. This can reduce the total number of transfers required to get funds to the real sender.

source/light/contracts/Light.sol
```diff
@@ -93,51 +104,65 @@ contract Light is ILight {
...
   function swap(
     uint256 nonce,
     uint256 expiry,
+    address signerWallet,
     IERC20 signerToken,
     uint256 signerAmount,
     IERC20 senderToken,
     uint256 senderAmount,
-    bytes calldata signature
+    uint8 v,
```
wjmelements, line 116: Optional: you can stash this uint8 in the high bits of an `address` to save 32 bytes of calldata.

source/light/contracts/Light.sol
```diff
@@ -146,6 +171,25 @@ contract Light is ILight {
     );
   }
 
+  function setFee(uint256 newFee) external onlyOwner {
+    fee = newFee;
```
wjmelements, line 175: require(fee < X)

vdrg: thanks, yes that is 100% supposed to be there 👍 will finish this PR today

source/light/contracts/Light.sol
vdrg: good idea

source/light/contracts/Light.sol
vdrg: thought about it but most of that calldata will be 0s and would make it a little bit harder to read

wjmelements:
Discussed with Don: We request that the fee is incorporated into the signature versioning, so that if it changes it will invalidate all of the signer orders. This is because a change to fee would be a material change to the true signer amount.

wjmelements:
> Discussed with Don: We request that the fee is incorporated into the signature versioning, so that if it changes it will invalidate all of the signer orders. This is because a change to fee would be a material change to the true signer amount.

He's now suggested a constant fee, which would save gas and fix the issue.

wjmelements (approved):
You could also save gas if the feeWallet was a constant too. Instead of configuring fee recipient in this contract, you should configure an owner in the feeWallet.

vdrg:
> Discussed with Don: We request that the fee is incorporated into the signature versioning, so that if it changes it will invalidate all of the signer orders. This is because a change to fee would be a material change to the true signer amount.

yes! I had proposed that too but now we are going for a constant fee so that's fixed 😃

source/light/contracts/Light.sol
```diff
@@ -68,10 +73,13 @@ contract Light is ILight {
...
-  // Mapping of signer addresses to an optionally set minimum valid nonce
-  mapping(address => uint256) public override signerMinimumNonce;
+  mapping(address => address) public override authorized;
 
-  constructor() public {
+  address public feeWallet;
+
+  constructor(address _feeWallet, uint256 _fee) public {
+    require(_feeWallet != address(0), "INVALID_FEE_WALLET");
+    require(0 < _fee && _fee < FEE_DIVISOR, "INVALID_FEE");
```
dmosites, line 82: We can remove `0 < _fee &&` since it'll evaluate to zero below anyway.

dmosites (approved):
LGTM with one comment.

wjmelements:
Can this merge? Eager to use.

vdrg:
> Can this merge? Eager to use.

heey sorry for the late reply, this will get audited soon 👍

PR #586, opened 2021-03-22
Optimize AIP 12

wjmelements:
#### Changes
* removeAll: Fix transfer amount
* removeAll: Loop through set in reverse to make removal SSTOREs cheaper
* cache supportedTokenList in more contexts to save on keccaks

Reviewers @dmosites

### DIFF
--- a/source/registry/contracts/Registry.sol
+++ b/source/registry/contracts/Registry.sol
@@ -30,15 +30,17 @@ contract Registry is Ownable {
   }
 
   function addTokens(address[] calldata tokenList) external {
+    EnumerableSet.AddressSet storage supportedTokenList =
+      supportedTokens[msg.sender];
     uint256 transferAmount = 0;
-    if (supportedTokens[msg.sender].length() == 0) {
+    if (supportedTokenList.length() == 0) {
       transferAmount = transferAmount.add(obligationCost);
     }
     for (uint256 i = 0; i < tokenList.length; i++) {
       address token = tokenList[i];
-      if (!supportedTokens[msg.sender].contains(token)) {
+      if (!supportedTokenList.contains(token)) {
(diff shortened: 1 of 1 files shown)


wjmelements:
Trying to get other optimizations into openzeppelin (we'll see about that lol) but I think this is ready for review.

wjmelements:
The openzeppelin optimizations:

* [EnumerableSet: Clear](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2607)
* [EnumerableSet: Remove Boundary Check in _at](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2606)
* [EnumerableSet: Skip SSTORE On Last Item](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2605)

If any of these aren't accepted you can build from a fork that cherry-picks them.

source/registry/contracts/Registry.sol
```diff
@@ -65,13 +69,14 @@ contract Registry is Ownable {
     EnumerableSet.AddressSet storage supportedTokenList =
       supportedTokens[msg.sender];
     uint256 length = supportedTokenList.length();
-    for (uint256 i = 0; i < length; i++) {
-      address token = supportedTokenList.at(0);
+    uint256 originalLength = length;
+    while (length --> 0) {
```
dmosites, line 73: Prefer `length-- > 0` for clarity.

wjmelements: worth noting: this behavior does not work in solidity 0.8 because it does not appreciate the underflow.

wjmelements: of course that can be worked around with `unchecked`

wjmelements:
outdated; see https://github.com/airswap/airswap-protocols/pull/607

PR #607, opened 2021-05-06
Upgrade and Remove All Tokens in Reverse Order

wjmelements:
#### Changes
* Caches AddressSet storage, saving linear amount of keccak map lookups
* Upgrades to latest solidity, 0.8.4
* Upgrades to latest openzeppelin-contracts, which needs solidity ^0.8.0
* Reversed removeAll order, fixing test and saving gas with https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2605
* Removes SafeMath, no-longer needed with solidity ^0.8.0

Reviewers @dmosites

### DIFF
--- a/source/registry/contracts/Registry.sol
+++ b/source/registry/contracts/Registry.sol
@@ -1,10 +1,11 @@
 //SPDX-License-Identifier: Unlicense
-pragma solidity ^0.7.6;
+pragma solidity ^0.8.0;
 pragma abicoder v2;
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
-import "@openzeppelin/contracts/math/SafeMath.sol";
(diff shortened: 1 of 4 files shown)


source/registry/contracts/Registry.sol
```diff
@@ -90,13 +91,15 @@ contract Registry {
     uint256 length = supportedTokenList.length();
     require(length > 0, "NO_TOKENS_TO_REMOVE");
     address[] memory tokenList = new address[](length);
-    for (uint256 i = 0; i < length; i++) {
-      address token = supportedTokenList.at(0);
+        
+    for (uint256 i = length; i > 0;) {
+      i--;
```
wjmelements, line 96: I tried doing the decrement in the condition like

```
for (uint256 i = length; i-- > 0;) {
  //
}
```

However the decrement didn't work inside the condition. Seems to be a bug with Solidity. So I do it inside the loop instead, which should be equivalent.

PR #626, opened 2021-07-07
Event declaration and function ordering for Registry.sol

dmosites:


wjmelements (approved):
I don't see the benefit of this. Linters are weird. It looks like the code is only moved within the contract and not otherwise changed.

PR #628, opened 2021-08-04
feature-update/metadata-fetchtokens

hmalik88:
## Description

The `fetchTokens` function in the metadata repo will currently fail if one of the urls it is fetching throws an error, because we are using `Promise.all`. 

Quick description:

- [] N/A
- [ ] New features
- [x] Bug fixes
- [x] Typo/small tweaks

## Changes

This PR changes the code from `Promise.all` -> `Promise.allSettled` and the return `Promise<Array<TokenInfo>>` -> `Promise<{tokens: TokenInfo[]; errors: string[]}>`

## Tests

Updated tests to accommodate the new return value of `fetchTokens`

### DIFF
--- a/tools/metadata/index.ts
+++ b/tools/metadata/index.ts
@@ -4,18 +4,34 @@ import { TokenInfo } from '@uniswap/token-lists'
 import { defaults, known } from './src/constants'
 import { getTokenName, getTokenSymbol, getTokenDecimals } from './src/helpers'
 
-export async function fetchTokens(chainId: number): Promise<Array<TokenInfo>> {
-  const tokens = await Promise.all(
+export async function fetchTokens(
+  chainId: number
+): Promise<{ tokens: TokenInfo[]; errors: string[] }> {
+  const errors = []
+  let tokens = []
+  const promises = await Promise.allSettled(
     known.map(async url => {
-      const { data } = await axios.get(url)
-      return data.tokens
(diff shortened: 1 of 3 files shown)


tools/metadata/index.ts
```diff
@@ -4,18 +4,34 @@ import { TokenInfo } from '@uniswap/token-lists'
...
+  let tokens = []
+  const promises = await Promise.allSettled(
     known.map(async url => {
-      const { data } = await axios.get(url)
-      return data.tokens
+      try {
+        const res = await axios.get(url)
+        return res.data.tokens
+      } catch (e) {
+        throw new Error(e.message)
```
dmosites, line 18: Is it necessary to throw a new Error with the same message?

hmalik88: Nope, it's unnecessary, just pushed up a commit removing it.

dmosites: Actually meant to ask whether the try / catch here is necessary. If `axios.get` throws will the encapsulating promise status not be `rejected`?

hmalik88: Ahh didn't realize the error bubbles to the encapsulating promise, okay pushing up a commit removing the try / catch

dmosites:
Please bump metadata package version to `0.3.3`, which will automatically publish it to NPM.

PR #662, opened 2021-09-04
@airswap/metadata: error handling

dmosites:
Return a `message` and `url` for errors; remove AST check in tests

### DIFF
--- a/tools/metadata/index.ts
+++ b/tools/metadata/index.ts
@@ -4,9 +4,6 @@ import { TokenInfo } from '@uniswap/token-lists'
 import { defaults, known } from './src/constants'
 import { getTokenName, getTokenSymbol, getTokenDecimals } from './src/helpers'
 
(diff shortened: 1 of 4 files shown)


hmalik88 (commented):
LGTM, however one of the tests is failing: 

Screen Shot 2021-09-04 at 11 54 37 AM

I had a sandbox from before that I was using to work on the metadata package and I saw this in the console when running your code: 

Screen Shot 2021-09-04 at 11 54 51 AM

Weird thing is that those URLs work fine in my browser, anyway those two are the lists are the ones that have WETH, so that explains the test failing.

dmosites:
Those servers must look at the request headers to decide whether to respond OK. Given that they've started erroring out in test environments I've removed the impacted mainnet tests and left the rinkeby ones so that we still test those functions. Also removed the shim. 👍

PR #672, opened 2021-09-20
Last look client

gpxl-dev:
Implementing last look for the client as per [specification here](https://about.airswap.io/technology/last-look).

All feedback appreciated.


tools/protocols/src/Server.ts
```diff
@@ -132,21 +159,168 @@ export class Server {
...
+  }
+
+  /**
+   * Throws if Last Look protocol is not supported, or if server is not yet
+   * initialized.
+   */
+  private requireLastLookSupport(version?: string) {
+    if (!this.supportsProtocol('last-look', version))
+      throw new Error(
+        `Server at ${this.locator} doesn't support this version of Last Look`
```
dmosites, line 320: Possible to conditionally throw either "Server at ... doesn't support Last Look" or if version is set "Server at ... doesn't support this version of Last Look"?

tools/protocols/src/Server.ts
```diff
@@ -1,69 +1,114 @@
...
+
+if (!isBrowser) {
+  JsonRpcWebsocket.setWebSocketFactory((url: string) => {
+    const ws = require('websocket').client
+    return new ws(url)
+  })
+}
+
+export class Server extends EventEmitter {
+  public transportProtocol: 'websockets' | 'http'
```
dmosites, line 50: Suggest singular `websocket`

dmosites (approved):
Looks good! Suggest we update the package name to `@airswap/libraries` (start with version `0.0.1`) and rename its folder from `protocols` to `libraries` as well. Just a couple other comments as well.

hmalik88 (approved):
LGTM

PR #681, opened 2021-09-25
Feature/wrapper

codyenokida:
Please let me know if I did this wrong!!


codyenokida:
Do the metadata test always fail? Or was it something that I implemented that caused this...

tools/libraries/src/Wrapper.ts
```diff
@@ -0,0 +1,62 @@
...
+      order.expiry,
+      order.signerWallet,
+      order.signerToken,
+      order.signerAmount,
+      order.senderToken,
+      order.senderAmount,
+      order.v,
+      order.r,
+      order.s,
+      { value: order.senderAmount }
```
dmosites, line 59: Transaction value should be 0 unless `order.senderToken` is WETH, in which case value is `order.senderAmount`

dmosites (changes requested):
Looks good, one comment!

dmosites (approved):

PR #684, opened 2021-09-27
Staking contract streamline balanceOf and delegate functionality

AboldUSER:


source/staking/test/unit.js
```diff
@@ -326,44 +214,30 @@ describe('Staking Unit', () => {
         await ethers.provider.send('evm_mine')
```
hmalik88, line 214: I know you didn't add this in, but I was looking and this is a really weird way to increase time...we're making a 100 `evm_mine` calls...this particular test is much slower than the others...you can change it to the following and it works faster:

`    await ethers.provider.send('evm_increaseTime', [10])`
`    await ethers.provider.send('evm_mine')`

source/staking/test/unit.js
```diff
@@ -376,210 +250,180 @@ describe('Staking Unit', () => {
...
```
hmalik88, line 250: Same deal here as the above.

source/staking/test/unit.js
```diff
@@ -376,210 +250,180 @@ describe('Staking Unit', () => {
...
-      await staking.connect(account1).stake('200')
-      const block0 = await ethers.provider.getBlock()
-      await staking.connect(account1).stake('300')
-      const block1 = await ethers.provider.getBlock()
-
-      // move 100 seconds forward + 2 stakes = 102% vested
-      for (let index = 0; index < 100; index++) {
+      // move 10 seconds forward - 10% vested
+      for (let index = 0; index < 10; index++) {
         await ethers.provider.send('evm_mine')
```
hmalik88, line 273: Same as above

source/staking/test/unit.js
```diff
@@ -376,210 +250,180 @@ describe('Staking Unit', () => {
...
-      await staking.connect(deployer).setVesting(DURATION, CLIFF)
-      await staking.connect(account1).stake('100')
+      await staking.connect(account1).addDelegate(account2.address)
+      await staking.connect(account2).stake('100')
 
-      const block = await ethers.provider.getBlock()
-      await ethers.provider.send('evm_mine', [block['timestamp'] + CLIFF])
+      // move 10 seconds forward - 10% vested
+      for (let index = 0; index < 10; index++) {
+        await ethers.provider.send('evm_mine')
```
hmalik88, line 388: Same as above.

source/staking/test/unit.js
```diff
@@ -376,210 +250,180 @@ describe('Staking Unit', () => {
...
+      await staking.connect(account1).addDelegate(account2.address)
+      await staking.connect(account2).stake('100')
+      await staking.connect(account2).stake('100')
+
+      const initialUserStake = await staking
+        .connect(account2)
+        .getStakes(account1.address)
+
+      // move 10 seconds forward - 10% vested
+      for (let index = 0; index < 10; index++) {
```
hmalik88, line 415: Same as above

hmalik88 (changes requested):
Apart from the things I commented on, LGTM.

source/staking/contracts/Staking.sol
```diff
@@ -85,57 +74,69 @@ contract Staking is Ownable {
...
-        _unstake(length, amounts[length]);
-        totalAmount += amounts[length];
-      }
-    }
+    address account;
+    isDelegate[msg.sender]
+      ? account = delegateAccount[msg.sender]
+      : account = msg.sender;
+    _unstake(account, amount);
+    totalAmount += amount;
```
dmosites, line 122: Think we just need to use `amount` here since there's only one.

source/staking/contracts/Staking.sol
```diff
@@ -85,57 +74,69 @@ contract Staking is Ownable {
...
-   * @param amount uint256
+   * @notice Add delegate for account
+   * @param delegate address
    */
-  function stake(uint256 amount) external {
-    stakeFor(msg.sender, amount);
+  function addDelegate(address delegate) external {
+    require(!isDelegate[delegate], "ALREADY_DELEGATE");
+    require(allStakes[delegate].balance == 0, "ALREADY_STAKING");
+    accountDelegate[msg.sender] = delegate;
```
dmosites, line 83: This will set `delegate` whether that account wishes to be set as a delegate. Recommend that we include some kind of propose-accept flow to protect from mistakes and potential abuse.

dmosites (changes requested):

source/staking/contracts/Staking.sol
```diff
@@ -16,21 +16,23 @@ contract Staking is Ownable {
...
+  uint256 public vestingLength;
 
   // Mapping of account to stakes
-  mapping(address => Stake[]) public allStakes;
+  mapping(address => Stake) internal allStakes;
+
+  // Mapping of account to delegate
+  mapping(address => address) public accountDelegate;
+  mapping(address => address) public delegateAccount;
+  mapping(address => bool) public isDelegate;
```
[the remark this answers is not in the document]
AboldUSER: Good call, implementing `if (delegateAccount[account] != address(0)) { ... }` works

source/staking/test/unit.js
AboldUSER: Done

source/staking/contracts/Staking.sol
AboldUSER: Created a new function called proposeDelegate, which is called by an account, and now setDelegate is called & confirmed by the proposed delegater

hmalik88 (approved):
LGTM 👍

dmosites (approved):

PR #686, opened 2021-10-03
Pool withdraw

AboldUSER:
Quick description: Add additional withdraw functions to allow for withdraw with recipient, withdraw and stake, and withdraw and stake for another account.


hmalik88 (approved):

dmosites:
Looking good though we should figure out how to test with a mock Staking contract rather than a copy of Staking.sol. If there are particular issues or errors being thrown let's discuss!

hmalik88:
> Looking good though we should figure out how to test with a mock Staking contract rather than a copy of Staking.sol. If there are particular issues or errors being thrown let's discuss!

@dmosites I think the issue is that the `stakeFor` call in `withdrawAndStake` and `withdrawAndStakeFor` is not succeeding. Not sure if its because the function itself doesn't return anything (so we can't do `.stakeFor.returns(true)`). We tried to throw in a `.transferFrom.returns(true)` to ensure the `transferFrom` in `stakeFor` succeeds, but no dice. I have a feeling it might be because the mock stake contract doesn't get initialized with the state variables that the `stakeFor` function refers to when being called. Not sure what it could be apart from that.

source/pool/contracts/Pool.sol
```diff
@@ -116,34 +174,105 @@ contract Pool is Ownable {
...
+   * @param account address
+   * @param recipient address
+   */
   function withdrawProtected(
     Claim[] memory claims,
     IERC20 token,
-    uint256 minimumAmount
-  ) public {
+    uint256 minimumAmount,
+    address account,
```
dmosites, line 252: Two thoughts to discuss in more detail:
1. If Bob has Alice's claims, could he call this function directly with her as the `account` and himself as the `recipient` and take her reward?
2. If `account` is always passed as `msg.sender` would it be most clear to just use `msg.sender` in this function as in the previous setup?

dmosites:
See that we're using `ethers.getContractFactory(STAKING.abi, STAKING.bytecode)` which should be fine for testing. I've removed the `test/Staking.sol` so think we're good to proceed after the discussion point above.

source/pool/contracts/Pool.sol
```diff
@@ -116,34 +174,105 @@ contract Pool is Ownable {
...
+  function withdrawAndStakeFor(
+    Claim[] memory claims,
+    IERC20 token,
+    address account
+  ) external {
+    require(token == IERC20(stakingToken), "INVALID_TOKEN");
+    uint256 amount = withdrawProtected(
+      claims,
+      token,
+      0,
```
dmosites, line 229: Should these be zero or be able to bet set by the client? Maybe `0` as a default?

source/pool/contracts/Pool.sol
AboldUSER: Yes to both of those points. I'll remove account and replace it with msg.sender in that withdrawProtected function.

source/pool/contracts/Pool.sol
AboldUSER: For the original withdraw function, it was coded 0 by default, so I carried that over to withdrawWithRecipient, withdrawAndStake, and withdrawAndStakeFor.
I can see that it doesn't allow for a client to set a minimum, so update that as you mention. 
I am thinking to allow this for all the withdraw functions, and then set withdrawProtected to internal, as I don't see why it would need to be called as the other withdraw functions cover the same functionality.

dmosites: The original `withdraw` was for backwards compatibility so had just set minimum to `0`. Normal usage will still be to call withdraw directly so do need a public version of it, even if shared logic is moved to an internal function.

AboldUSER: - Updated the new withdraw functions (withdrawWithRecipient, withdrawAndStake, withdrawAndStakeFor) with a minimumAmount parameter
- Kept the original withdraw function as is for backwards compatibility and maintained the withdrawProtected function as public

dmosites (approved):

PR #689, opened 2021-10-07
Initial wrapper updates and integration test

dmosites:


AboldUSER:
Overall looks good. Only comment would be if we should have a function to update the light contract address in case it changes, so we don't need to deploy a new wrapper.

PR #692, opened 2021-10-08
Initial light rebates and integration test

dmosites:


AboldUSER:
Would this revoke function in Light.sol ever need to have the address to be revoked passed as an argument, instead of always being the message sender?

```
function revoke() external override {
    address tmp = authorized[msg.sender];
    delete authorized[msg.sender];
    emit Revoke(tmp, msg.sender);
  }
```

PR #718, opened 2021-11-03
Renames light to swap and cleanup

dmosites:


AboldUSER (commented):
Went through testing and everything passes as both unit and integration tests. 
Reviewing the word changes and clean up around light to swap makes sense.
The only thing I suggest is remove the * in *unit.js in the package.json in swap.

PR #734, opened 2021-12-01
Fix wrapper swapContract setter, add sellNFT and buyNFT to wrapper

AboldUSER:
Fixed setSwapContract to zero out allowance when new address is set.
Added sellNFT and buyNFT functions to wrapper contract.

### DIFF
--- a/source/swap/contracts/Swap.sol
+++ b/source/swap/contracts/Swap.sol
@@ -691,7 +691,7 @@ contract Swap is ISwap, Ownable {
     uint256 feeAmount = amount.mul(protocolFee).div(FEE_DIVISOR);
     if (feeAmount > 0) {
       uint256 discountAmount = calculateDiscount(
-        IERC20(stakingToken).balanceOf(sourceWallet),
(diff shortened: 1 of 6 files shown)


source/swap/contracts/Swap.sol
```diff
@@ -676,6 +676,29 @@ contract Swap is ISwap, Ownable {
     return rebateMax.mul(stakingBalance).mul(feeAmount).div(divisor).div(100);
   }
 
+  /**
+   * @notice Calculates and transfers protocol fee and rebate
```
dmosites, line 680: "Calculates and returns fee amount"

source/swap/contracts/Swap.sol
```diff
@@ -676,6 +676,29 @@ contract Swap is ISwap, Ownable {
...
+   * @param sourceWallet address
```
dmosites, line 681: `wallet`

dmosites (approved):
Looks good with two suggestions

PR #739, opened 2021-12-14
Add value & volume and correct fees for subgraph

gpxl-dev:
Adds `swapValueUsd` and `feeValueUsd` to `SwapLight` entity and adds an additional `VolumeDay` entity.

This PR is largely based on existing code to calculate fees, with a correction to a small but significant error in the calculation.

I had a question as to whether `computeValueUsd` should actually be returning a BigDecimal, and the same with `computeFeeAmountUsd` - would like to test to see whether sub $1 fees show as 0.

I changed the approach from manually listing the chainlink price oracles in `pricingHelper` to instead using the [Feed Registry](https://docs.chain.link/docs/feed-registry/). I can see that this was only [deployed on July 20th](https://etherscan.io/tx/0x9acce3b01d128f97508917a81c2f1a8aa934872617b0f8a4970bf9cba68e5354) so I'm wondering if this means our subgraph will only show fees from then, and whether that matters?

I also noticed that the subgraph had been upgraded to use `protocolFee` instead of `signerFee` - this was possibly a repo-wide find and replace - AFAIK this change is not live yet, so am thinking perhaps we should revert that for now?

Finally worth noting there's more work to do on this subgraph - we can tidy it up, but also need to consider adding support for the wrapper and probably other new features.


AboldUSER (approved):
This looks good to me, so will approving it.

Issue #736, opened 2021-12-06
Create a new pool function to accept individually signed claims


dmosites:

francisldn:
To do:
1. withdrawWithSignature should calculate amount from points rather than directly transfer the passed amount.
2. we should use nonce to mark claims as used. the Activate backend will issue claims with the same nonce until a Claim event is confirmed. then it'll issue claims with nonce + 1.
3. we can remove all other logic related to merkle trees and roots and limit to individually signed claims.
4. To use EIP712 for signing

PR #743, opened 2021-12-17
added withdrawWithSignature function

francisldn:
Draft PR for the issue - Create a new pool function to accept individually signed claims #736

### DIFF
--- a/test/unit.js
+++ b/test/unit.js
@@ -6,7 +6,7 @@ const { generateTreeFromData, getRoot, getProof } = require('@airswap/merkle')
 const { ethers, waffle } = require('hardhat')
 const { deployMockContract } = waffle
 const IERC20 = require('@openzeppelin/contracts/build/contracts/IERC20.json')
-const STAKING = require('../build/contracts/Staking.sol/Staking.json')
+const STAKING = require('@airswap/staking/build/contracts/Staking.sol/Staking.json')
 
 function toWei(value, places) {
   return toAtomicString(value, places || 18)
--- a/test/unit.js
+++ b/test/unit.js
@@ -1,0 +1,658 @@
+const { expect } = require('chai')
+const { soliditySha3 } = require('web3-utils')
+const { toAtomicString } = require('@airswap/utils')
+const { generateTreeFromData, getRoot, getProof } = require('@airswap/merkle')
+
+const { ethers, waffle } = require('hardhat')
+const { deployMockContract } = waffle
+const IERC20 = require('@openzeppelin/contracts/build/contracts/IERC20.json')
+const STAKING = require('../build/contracts/Staking.sol/Staking.json')
+
+function toWei(value, places) {
+  return toAtomicString(value, places || 18)
+}
+
+describe('Pool Unit Tests', () => {
+  let deployer
+  let alice
+  let bob
+  let carol
+  let stakeContract
+
+  const CLAIM_SCALE = 10
+  const CLAIM_MAX = 50
+
+  const ALICE_SCORE = toWei(10000, 4)
+  const BOB_SCORE = toWei(100000, 4)
+  const CAROL_SCORE = toWei(1000000, 4)
+
+  let tree
+  let feeToken
+  let feeToken2
+  let pool
+  let snapshotId
+
+  beforeEach(async () => {
+    snapshotId = await ethers.provider.send('evm_snapshot')
+  })
+
+  afterEach(async () => {
+    await ethers.provider.send('evm_revert', [snapshotId])
+  })
+
+  before(async () => {
+    ;[deployer, alice, bob, carol] = await ethers.getSigners()
+
+    feeToken = await deployMockContract(deployer, IERC20.abi)
+    await feeToken.mock.approve.returns(true)
+    feeToken2 = await deployMockContract(deployer, IERC20.abi)
+
+    stakeContract = await (
+      await ethers.getContractFactory(STAKING.abi, STAKING.bytecode)
+    ).deploy(feeToken.address, 'StakedAST', 'sAST', 100, 10)
+    await stakeContract.deployed()
+
+    pool = await (
+      await ethers.getContractFactory('Pool')
+    ).deploy(CLAIM_SCALE, CLAIM_MAX, stakeContract.address, feeToken.address)
+    await pool.deployed()
+
+    tree = generateTreeFromData({
+      [alice.address]: ALICE_SCORE,
+      [bob.address]: BOB_SCORE,
+      [carol.address]: CAROL_SCORE,
+    })
+  })
+
+  describe('Test constructor', async () => {
+    it('constructor sets values', async () => {
+      const storedScale = await pool.scale()
+      const storedMax = await pool.max()
+      const stakingContract = await pool.stakingContract()
+      const stakingToken = await pool.stakingToken()
+      const adminOwner = await pool.admins(deployer.address)
+      expect(storedScale).to.equal(CLAIM_SCALE)
+      expect(storedMax).to.equal(CLAIM_MAX)
+      expect(stakingContract).to.equal(stakeContract.address)
+      expect(stakingToken).to.equal(feeToken.address)
+      expect(adminOwner).to.equal(true)
+    })
+
+    it('constructor reverts when percentage is too high', async () => {
+      await expect(
+        (
+          await ethers.getContractFactory('Pool')
+        ).deploy(CLAIM_SCALE, 101, stakeContract.address, feeToken.address)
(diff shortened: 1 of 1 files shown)


francisldn:
@dmosites following the discussion yesterday, I thought that instead of recovering the signer, we can just compare the signed hash vs the hash provided by Activate. If the hashes match, then the token amount will be transferred.

AboldUSER:
Awesome and a couple comments:

- The modifier MultiAdmin can be used instead of the require function, to align with similar use of the admins mapping in the contract

- I see a second mapping is created, hashClaimed. If we have two different claim mappings, is there a risk that a user can receive double claim - once from the merkle tree route and second from an admin calling this function? Or if the signed message route is for claims not on the merkle tree then never mind

-  This piece of code (hashClaimed[ethSignedMessageHash] = true;) can go before the safeTransfer call. I think it is just always good practice to avoid putting state changing pieces after calling an external address

- In the current function code, I don't think it is set up to take a signature, instead it is taking the hashed data portion that would normally then be signed. So in this case Activate or whoever the admin is wont be sending a signed messaged, instead just sending token, recipient and amount arguments and a hash of those arguments. If Activate is indeed providing a signed message, then we'd have to use the ECDSA.recover function on the getEthSignedMessageHash output. In the case that they are just sending hashed data, I don't see the point of sending the arguments and then the hash of those arguments, and then verify the arguments and hash in the contract as it is just redundant. 

- The withdrawWithSignature can only ever be called on a recipient, amount, and token once. Since we are mapping the hash of those arguments to a bool, we would never be able to claim that same token & amount for a recipient in the future. Not sure if this is expected, but otherwise then the signed message would need to also contain at least salt/nonce.

- I don't think true needs to be returned, as there is also an event being emitted. Maybe can return an amount value similar to the withdrawProtected function

francisldn:
Thank you @AboldUSER. All very valid points. I have made some changes to the function based on your feedback:
- used the same "claimed" mapping for withdrawWithSignature 
- added nonce to the hash - assuming nonce is recorded off-chain
- added signature as an argument
- used multiAdmin modifier
- state change before external call
- return amount (uint) instead of boolean

francisldn:
2 further changes:
- updated the IPool interface - adding withdrawWithSignature function
- added 2 additional tests for the withdrawWithSignature function in the integration.js file

source/pool/contracts/Pool.sol
```diff
@@ -363,4 +368,34 @@ contract Pool is IPool, Ownable {
...
+  * @param amount uint256
+  * @param recipient address
+  */
+  function withdrawWithSignature(
+    bytes memory signature,
+    bytes32 messageHash,
+    address token,
+    uint256 amount,
+    address recipient
+  ) external override multiAdmin returns (uint256) {
```
dmosites, line 385: Should be callable by anyone. Admins sign and distribute claims to recipients who can in turn call this function to withdraw.

source/pool/contracts/Pool.sol
```diff
@@ -363,4 +368,34 @@ contract Pool is IPool, Ownable {
...
+    // verify signed hash has not been claimed
+    require(!claimed[messageHash][recipient], "CLAIM_ALREADY_MADE");
+    // to verify hash is signed by an admin, ECDSA.recover will throw error if signer is not recoverable
+    bytes32 ethSignedMessageHash = ECDSA.toEthSignedMessageHash(messageHash);
```
dmosites, line 389: Hash should be calculated inline from the token, amount, recipient, and a nonce. Otherwise, anyone with a valid hash and signature can transfer any amount of any token.

source/pool/contracts/Pool.sol
```diff
@@ -363,4 +368,34 @@ contract Pool is IPool, Ownable {
...
+    address signer = ECDSA.recover(ethSignedMessageHash, signature);
+    require(signer == msg.sender, "NOT_VERIFIED");
```
dmosites, line 391: This should require that signer is among the admins i.e. `admins[signer] == true`

source/pool/contracts/Pool.sol
```diff
@@ -363,4 +368,34 @@ contract Pool is IPool, Ownable {
...
+    // mark the hash of (signature, hash and nonce) and recipient as true
+    claimed[messageHash][recipient] = true;
```
dmosites, line 393: If hash is calculated inline (including recipient, token, amount, nonce) as suggested above, this just needs to be `claimed[hash] = true`

source/pool/contracts/Pool.sol
```diff
@@ -363,4 +368,34 @@ contract Pool is IPool, Ownable {
...
+/** @notice withdraw function that uses signature instead of claim
+  * @param signature signature from activate or any admin
+  * @param messageHash hash of token address, minimumAmount, recipient and nonce
+  * @param token address
+  * @param amount uint256
+  * @param recipient address
+  */
+  function withdrawWithSignature(
+    bytes memory signature,
+    bytes32 messageHash,
```
dmosites, line 381: Should take a `nonce` argument and calculate the hash in the function body (see below).

dmosites (changes requested):

source/pool/contracts/Pool.sol
francisldn: Hi @dmosites thanks for the comments, sorry I misunderstood the requirements. I have a further question, in order to recover the signer, we need a signed hash and a signature of one of the admins, should "signed hash" be one of the arguments of the function? Does activate produce a signed hash?

dmosites: An admin produces an off-chain claim that includes claim parameters and a signature. The `ecrecover` solidity function takes a hash and signature parameters as arguments. Should be able to compute a hash from claim parameters at runtime similar to what's happening in Swap: https://github.com/airswap/airswap-protocols/blob/main/source/swap/contracts/Swap.sol#L208

francisldn: I have modified the function so that it can now recover the signer from the hash and signature using ``ECDSA.recover()`` function. I didn't use ``ecrecover`` function because it allows for malleable (non-unique) signatures according to Openzeppelin.

source/pool/contracts/Pool.sol
francisldn: ok this is modified accordingly.

source/pool/contracts/Pool.sol
francisldn: ok, this is modified accordingly.

source/pool/contracts/Pool.sol
francisldn: ``nonce`` added to the argument.

dmosites: Do we need the `messageHash` argument if we're generating it in the function body?

francisldn: We would need it for verification purpose? If a user enters the wrong amount or token address (intentionally or unintentionally) then the generated hash will not match with the ``messageHash``.

AboldUSER: Generally, signed message implementations take the individual parameters and the signature as arguments. The signature contains the message hash anyways. 
If a user is inputting incorrect arguments in a function, then you would want the transaction to fail.

francisldn: Ok, so I will remove the ``messageHash`` and ``require(messageHash == calculatedHash, "HASH_NOT_VERIFIED")`` as any wrong input will result in the ``ECDSA.recover`` transaction to fail, so there is no need to compare the hash. Am I understanding this correctly?

AboldUSER: Yup, that is correct.

source/pool/contracts/Pool.sol
```diff
@@ -363,4 +368,41 @@ contract Pool is IPool, Ownable {
...
+  */
+  function withdrawWithSignature(
+    uint8 v,
+    bytes32 r,
+    bytes32 s,
+    bytes32 messageHash,
+    address token,
+    uint256 amount,
+    uint256 nonce
+  ) external override returns (uint256) {
```
AboldUSER, line 389: Since this is an external, state-changing function, it should go above internal and/or pure/view functions as a styling best practice. I'd suggest moving it to just above the withdrawProtected function.

source/pool/test/integration.js
```diff
@@ -293,8 +294,58 @@ describe('Pool Integration Tests', () => {
         )
       ).to.be.revertedWith('INVALID_TOKEN')
     })
-  })
 
+    it('withdrawWithSignature does not revert with signer who is admin', async () => {
```
AboldUSER, line 298: I think these should actually go in the unit test file, as these functions are not interacting with any other contract.

PR #755, opened 2022-01-07
Converter swap coverage

AboldUSER:
Added additional tests to Converter contract unit test file and Swap contract unit test file.
Converter has 100% full test coverage.
Swap has 97.22% branch test coverage as a require statement is present that checks chain ID during signature verification in the event of a fork and change of chain ID. Was not able to find a way to configure change of chain ID within a test.


dmosites:
Looks good thanks @AboldUSER!

PR #775, opened 2022-01-27
Update withdraw event index to Pool contract

AboldUSER:
Indexed the nonce and account parameters within the Withdraw event within the Pool contract.
This supports off-chain monitoring of pool transactions/events.


AboldUSER:
This is redundant now.

Issue #776, opened 2022-01-28
PVE001: Proper Allowance Reset for Old Staking Contracts


dmosites:
![photo_2022-01-28 13 23 08](https://user-images.githubusercontent.com/1601469/151546442-54a65f48-e7ff-4c7d-acb6-7c0ce4f24dd0.jpeg)

AboldUSER:
This would require calling approve from the staking token with the arguments of the old staking contract and zero

PR #786, opened 2022-01-29
Reset allowance on old stakingToken and stakingContract to zero #776

francisldn:
Audit item PVE001 - Proper Allowance Reset for Old Staking Contracts #776


dmosites (approved):
LGTM

PR #797, opened 2022-02-02
Fix swap staking balanceOf check to staking contract

AboldUSER:
Updated the fee rebate checker to check balanceOf within the staking contract, instead of just the staking token.
This will ensure senders cannot utilize flashloans to add the staking token to their account, as they would have to stake in the staking contract for a certain duration.


francisldn (approved):
Looks good.

PR #798, opened 2022-02-03
Update approval to safeApproval and safeIncreaseAllowance

AboldUSER:
Updated pool contract to use safeApprove. The reasoning to use this function instead of safeIncreaseAllowance & safeDecreaseAllowance is because we are only initializing an allowance to max approval value or resetting an existing allowance to zero, which aligns with the appropriate usage in the OpenZeppelin docs. Also, if we used safeDecreaseAllowance, then there may need to be an additional external call to get the current allowance.

Updated the converter contract to use safeIncreaseAllowance. The reasoning is that we never decrease the allowance, and even though the allowance should go back zero after going through the AMM doing this just to be extra safe.

### DIFF
--- a/source/converter/contracts/Converter.sol
+++ b/source/converter/contracts/Converter.sol
@@ -128,7 +128,7 @@ contract Converter is Ownable, ReentrancyGuard, TokenPaymentSplitter {
         path = getTokenPath(_swapFromToken);
       }
       // Approve token for AMM usage.
-      IERC20(_swapFromToken).safeIncreaseAllowance(address(uniRouter), tokenBalance);
+      IERC20(_swapFromToken).safeIncreaseAllowance(uniRouter, tokenBalance);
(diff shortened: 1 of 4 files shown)


source/converter/contracts/Converter.sol
```diff
@@ -128,7 +128,7 @@ contract Converter is Ownable, ReentrancyGuard, TokenPaymentSplitter {
         path = getTokenPath(_swapFromToken);
       }
       // Approve token for AMM usage.
-      _approveErc20(_swapFromToken, tokenBalance);
+      IERC20(_swapFromToken).safeIncreaseAllowance(address(uniRouter), tokenBalance);
```
francisldn, line 131: ``uniRouter`` is declared as address type, why is there a need to cast it to 'address' again?

francisldn (commented):
Just one question on ``uniRouter``, the rest looks good.

francisldn (approved):

PR #801, opened 2022-02-09
Add expiry to pool claim

AboldUSER:
Updated pool contract, tests, and repo's typescript/utils to include expiry in pool claims.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6428,9 +6428,9 @@ globby@^11.0.2, globby@^11.0.4:
     merge2 "^1.3.0"
     slash "^3.0.0"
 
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -6428,9 +6428,9 @@ globby@^11.0.2, globby@^11.0.4:
     merge2 "^1.3.0"
     slash "^3.0.0"
 
-"gluegun@git+https://github.com/edgeandnode/gluegun.git#v4.3.1-pin-colors-dep":
```
francisldn, line 6431: Is this to fix the error in yarn lock file?

AboldUSER: Yes, when I pulled and did yarn install, I noticed the yarn lock file updated with this, so I committed it to see if it'll fix the current CI issue.

francisldn (approved):

PR #831, opened 2022-03-21
Swap deploy

dmosites:
Deploys for swap and wrapper contracts. Necessary due to an incorrect revert reason in `swap.setStaking` that was fixed. For mainnet and avalanche used `wrapper.setSwapContract` rather than redeploying.


wjmelements:
> For mainnet and avalanche used `wrapper.setSwapContract` rather than redeploying.

What does this mean?

source/wrapper/deploys.js
```diff
@@ -1,11 +1,11 @@
 module.exports = {
   1: '0x3a0e257568cc9c6c5d767d5dc0cd8a9ac69cc3ae',
```
wjmelements, line 2: ok I see; you don't deploy a new wrapper.

Ideally your wrapper doesn't need to SLOAD the swap address.

If you're only using storage for a rarely-changing configuration, you can consider moving that configuration to a separate contract, or making them immutable and using the reincarnation create2 upgrade pattern.

dmosites: Sounds good, will definitely look into this for a future wrapper contract. Thanks @wjmelements

PR #848, opened 2022-04-20
v3 subgraph files

geewynn:
Draft PR for airswap V3.

Using default schema and indexing all airswap protocol events emitted.

### DIFF
--- a/tools/subgraph-v3/src/PoolMapping.ts
+++ b/tools/subgraph-v3/src/PoolMapping.ts
@@ -1,0 +1,63 @@
+import {
+  DrainTo as DrainToEvent,
+  OwnershipTransferred as OwnershipTransferredEvent,
+  SetMax as SetMaxEvent,
+  SetScale as SetScaleEvent,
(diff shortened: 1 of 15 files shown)


tools/subgraph-v3/src/PoolMapping.ts
```diff
@@ -0,0 +1,63 @@
...
+  entity.tokens = event.params.tokens
+  entity.dest = event.params.dest
+  entity.save()
+}
+
+export function handleOwnershipTransferred(
+  event: OwnershipTransferredEvent
+): void {
+  let entity = new OwnershipTransferred(
+    event.transaction.hash.toHex() + "-" + event.logIndex.toString()
```
Mitsichury, line 29: Don't know if every event can be downcasted to a basic event type, if it is possible, it could be interesting to extract a formatting method 
let entity = new OwnershipTransferred(formatId(event))

dmosites:
Moved to https://github.com/airswap/airswap-subgraph

PR #856, opened 2022-06-29
Adds order hashing to utils

dmosites:
Usage:
```
import { getOrderHash } from '@airswap/utils';
getOrderHash(unsignedOrder, swapContract).toString('base64');
```


piersss (changes requested):
Small suggestion. No clue about the hashing stuff though. 😬

gpxl-dev (approved):

PR #858, opened 2022-06-30
Adds util functions for order param compression

dmosites:


tools/utils/test/utils.ts
```diff
@@ -95,4 +101,25 @@ describe('Utils', async () => {
...
+      ADDRESS_ZERO,
+      1,
+      v,
+      r,
+      s
+    )
+    const compressed = compressOrder({ ...unsignedOrder, v, r, s })
+    const signedOrder = decompressOrder(compressed)
+    expect(isValidOrder(signedOrder)).to.equal(true)
+    expect(signerWallet.toLowerCase()).to.equal(wallet.address.toLowerCase())
```
gpxl-dev, line 123: Could do:

```typescript
expect(signedOrder).to.deep.equal({ ...unsignedOrder, v, r, s})
```

gpxl-dev (approved):

tools/utils/test/utils.ts
dmosites: This would fail because `signedOrder` (type `Order`) does not include `protocolFee` or `senderWallet` as they are respectively compared to the `protocolFee` on the Swap contract and `msg.sender` during transaction execution.

However, compressed order strings could include these additional params (and others) if necessary for UX, even if they're not necessary to actually take the order.

PR #870, opened 2022-07-25
Swap Improvements

dmosites:


Perelyn-sama (commented):
Hey Don, I don't get why you changed the second argument of `safeTransferFrom` on line 227(Swap.sol) from  `msg.sender` to `recipent`. Since the recipient is expected to be the null address for public swaps and the plan to substitute `msg.sender` for it.

source/swap/contracts/Swap.sol
```diff
@@ -8,6 +8,8 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
 import "./interfaces/ISwap.sol";
 
+import "hardhat/console.sol";
```
dmosites, line 11: Please remove this and the related lines below @Perelyn-sama

dmosites:
> Hey Don, I don't get why you changed the second argument of `safeTransferFrom` on line 227(Swap.sol) from `msg.sender` to `recipent`. Since the recipient is expected to be the null address for public swaps and the plan to substitute `msg.sender` for it.

The aim is for the sender to be able to transfer their proceeds to an arbitrary `recipient`. In most cases the sender will just use their own wallet address for the `recipient` argument.

source/swap/contracts/Swap.sol
Perelyn-sama: On it, sorry about that.

Perelyn-sama:
> > Hey Don, I don't get why you changed the second argument of `safeTransferFrom` on line 227(Swap.sol) from `msg.sender` to `recipent`. Since the recipient is expected to be the null address for public swaps and the plan to substitute `msg.sender` for it.
> 
> The aim is for the sender to be able to transfer their proceeds to an arbitrary `recipient`. In most cases the sender will just use their own wallet address for the `recipient` argument.

Okay, I guess that makes sense. Thanks for clarifying.

source/swap/contracts/Swap.sol
```diff
@@ -787,13 +693,15 @@ contract Swap is ISwap, Ownable {
...
-      msg.sender,
+      senderWallet,
       senderToken,
       senderAmount
     );
 
     // Recover the signatory from the hash and signature
     address signatory = _getSignatory(hashed, v, r, s);
+    // console.log(signatory);
+    // console.log(signerWallet);
```
dmosites, line 704: Sorry should have flagged these as well @Perelyn-sama. Thanks.

PR #878, opened 2022-07-27
Indexer registry and goerli deploy

dmosites:


source/indexers/contracts/Indexers.sol
```diff
@@ -0,0 +1,48 @@
...
+
+  event Stake(address indexed account);
+  event Unstake(address indexed account);
+  event SetURL(address indexed account, string url);
+
+  /**
+   * @notice Set the URL for a staker
+   * @param _url string value of the URL
+   */
+  function setURL(string calldata _url) external {
```
Mitsichury, line 28: Optionnal:
It adds an URL to the set, not setting one 
Proposal :   function addURL(string calldata _url) external {...}

Mitsichury (approved):

source/indexers/contracts/Indexers.sol
dmosites: Good thought, though `setURL` "sets" the URL for `msg.sender` and there is symmetric no `removeURL` as the sender would just `setURL` with a blank string.

PR #888, opened 2022-08-28
feat(FullOrder): Add validator 

Mitsichury:


tools/utils/index.ts
```diff
@@ -131,6 +132,27 @@ export function isValidOrder(order: Order): boolean {
   )
 }
 
+export function isValidFullOrder(order: any | undefined | null): boolean {
```
GiuntaLucas, line 135: Why to check every props on an any object instead of a proper type and use return order != null (cover null and undefined) && order instanceof "MyOrderType" ?

Mitsichury: Javascript can't ensure a type validation at runtime.
Validator should ensure that an object can be cast or mapped as expected.
Would be safe to use Instance of in java or other strongly typed because we can't allocate an object that does not fit the class.

dmosites: Think it'd still be best for the order to be typed and its property requirements enforced at compile time even if there could be issues with the value at runtime e.g. `isValidFullOrder(order: FullOrder)` similar to existing `isValidOrder(order: Order)`

Mitsichury:
Repushed on this repo instead of the fork

PR #889, opened 2022-08-29
feat(FullOrder): Add validator

Mitsichury:


dmosites (changes requested):
Just need to bump the `@airswap/utils` package version from `3.0.3` to `3.0.4` so that it is (automatically) published to NPM.

dmosites (approved):

PR #899, opened 2022-09-08
Swap audit 3.2

Perelyn-sama:
Changes 
- Added TSOwnable to utils folder
- mported TSOwnable to the Swap.sol and made the Swap contract Inherit it.
- Updated unit tests, some tests were expecting the string 'Ownable: caller is not the owner' and had to change it to 'OnlyCallableByOwner()' since TSOwnable implemented a custom error.

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -77,7 +77,7 @@ module.exports = {
   solidity: {
     compilers: [
       {
-        version: '0.8.7',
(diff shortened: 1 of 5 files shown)


dmosites:
Prefer to use the audited `Ownable` and risk of mis-transferring ownership relatively low so going to close this for the current version of the swap contract.

PR #901, opened 2022-09-09
Additional order signer validity check

dmosites:


Perelyn-sama (approved):
Should we try a test to try it out?

Perelyn-sama:
I just found out the Authorize function already had a check for the zero address, does that make everything we did here pointless? 
@dmosites

PR #923, opened 2022-10-25
Publish latest changes to NPM

dmosites:
* Update registry names; update libraries names
* Naming update and bump libraries package version


piersss (approved):
Seems like just a lot of renaming, LGTM

PR #934, opened 2022-11-26
chore(indexer): Downgrade typing error to ES5

Mitsichury:


tools/libraries/package.json
```diff
@@ -31,7 +31,6 @@
     "@airswap/constants": "3.0.1",
     "@airswap/jsonrpc-client-websocket": "0.0.1",
     "@airswap/maker-registry": "3.1.2",
-    "@airswap/indexer-registry": "3.1.2",
```
dmosites, line 34: `@airswap/indexer-registry` is required by `src/IndexerRegistry.ts`

dmosites (changes requested):

dmosites (approved):

PR #940, opened 2022-11-30
Swap any to swap

smartcontrart:
Fixes issue #937


tools/utils/src/swap.ts
```diff
@@ -1,20 +1,21 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
dmosites, line 1: Please verify this is required

dmosites (changes requested):
Looks good two small comments.

dmosites (approved):

PR #955, opened 2022-12-13
Protocol fee update

smartcontrart:
Adjusting protocol fees to match swap-erc20
Fixes issue #928


dmosites (approved):
Looking good. Have a couple updates coming in another PR.

PR #969, opened 2022-12-23
Scrape default provider

dmosites:


tools/metadata/index.ts
```diff
@@ -52,12 +52,10 @@ export async function scrapeToken(
   let provider
   if (typeof ethersProvider === 'string') {
     provider = new ethers.providers.JsonRpcProvider(ethersProvider)
-  } else if (typeof ethersProvider !== undefined) {
+  } else if (ethersProvider !== null) {
```
piersss, line 55:
```suggestion
  } else if (!ethersProvider) {
```

Because it could also be undefined.

piersss (changes requested):

piersss (approved):

PR #971, opened 2022-12-31
Add erc777handler

### LINKED ISSUE #965: Add an ERC777 TransferHandler for new Swap

smartcontrart:
Fixes #965

### DIFF
--- a/source/swap/test/unit.js
+++ b/source/swap/test/unit.js
@@ -155,10 +155,10 @@ describe('Swap Unit Tests', () => {
       tokenKinds.ERC20,
       erc20Handler.address
     )
-    // await transferHandlerRegistry.addTransferHandler(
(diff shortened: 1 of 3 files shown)


source/swap/test/unit.js
```diff
@@ -141,6 +155,10 @@ describe('Swap Unit Tests', () => {
       tokenKinds.ERC20,
       erc20Handler.address
     )
+    // await transferHandlerRegistry.addTransferHandler(
```
dmosites, line 158: How does check work if the handler isn't added?

smartcontrart: The tests are wrong, I pushed an update to add the handler back but we need to review the tests. We have a similar issue with the unit test of erc1155.

dmosites (approved):

PR #978, opened 2023-01-10
Include staking token addresses in metadata

dmosites:


tools/metadata/index.ts
```diff
@@ -38,6 +38,15 @@ export async function fetchTokens(
...
     return token.address && token.chainId === chainId
   })
+  if (stakingTokenAddresses[chainId]) {
+    tokens.push({
+      name: 'Staked AST',
+      address: stakingTokenAddresses[chainId],
+      decimals: 4,
+      symbol: 'sAST',
+      chainId,
+    })
```
piersss, line 48:
```suggestion
    const tokens = Object.values(stakingTokenAddresses[chainId])
      .map((address) => getTokenInfo(address, tokens));
    tokens.push(tokens);
```

dmosites: > Something like this:

Since `fetchTokens` takes `chainId` isn't it just the one token that matters?

piersss: Let me paste the question I asked you 4 weeks ago:

> why doesn't metadata provide me with all staker tokens when using fetchTokens(chainId: number)?
> Because right now you get staker tokens from registry, then the tokens from fetchTokens and then any staker tokens that's not in fetchTokens gets called from scrapeToken.

Rephrased:

If `fetchTokens` gives me all the `TokenInfo` of the staker tokens, then I don't need to individually call the >10 tokens with `scrapeToken` which fails for the majority on testnet with a 429 error.

dmosites: There's one staking token per chain. Since `fetchTokens` takes a `chainId` it returns one staking token. Because staking tokens are the same except for address and chainId, how about this for the web app:
```
import { stakingTokenAddresses } from '@airswap/constants'
const stakingTokens = []
for (const chainId in stakingTokenAddresses) {
  stakingTokens.push({
    name: 'Staked AST',
    symbol: 'sAST',
    address: stakingTokenAddresses[chainId],
    decimals: 4,
    chainId
  })
}
```
Keep in mind though that the only staking token that matters is on mainnet / chainId = 1.

piersss: I'd like to have the TokenInfo of the `stakingTokenAddresses`, not just the addresses. Is that possible?

dmosites: Sure, added `stakingTokens` for import like so:
```
import { stakingTokens } from '@airswap/constants'

[
  {
    name: 'AirSwap Token',
    symbol: 'AST',
    address: '0x27054b13b1b798b345b591a4d22e6562d47ea75a',
    decimals: 4,
    chainId: 1
  },
  {
    name: 'AirSwap Token (GOERLI Placeholder)',
    symbol: 'AST (GOERLI Placeholder)',
    address: '0x1a1ec25dc08e98e5e93f1104b5e5cdd298707d31',
    decimals: 4,
    chainId: 5
  }, ...
]
```
FYI updated `chainId` to be number since that's what we get from ethers.

piersss (approved):
LGTM

PR #979, opened 2023-01-18
First wave of audit remediation recommendations

smartcontrart:
Pending integration of approval mechanism

### DIFF
--- a/source/swap/contracts/Swap.sol
+++ b/source/swap/contracts/Swap.sol
@@ -427,6 +427,11 @@ contract Swap is ISwap, Ownable {
       }
     }
 
+    if (order.protocolFee != protocolFee) {
+      errors[errCount] = "InvalidFee";
+      errCount++;
+    }
+
     return (errors, errCount);
   }
 
--- a/source/swap/contracts/Swap.sol
+++ b/source/swap/contracts/Swap.sol
@@ -127,8 +127,7 @@ contract Swap is ISwap, Ownable {
       revert NonceTooLow();
 
     // Ensure the nonce is not yet used and if not mark it used
-    if (!_markNonceAsUsed(order.signer.wallet, order.nonce))
-      revert NonceAlreadyUsed(order.nonce);
+    _markNonceAsUsed(order.signer.wallet, order.nonce);
 
     // Validate the sender side of the trade.
     address finalSenderWallet;
@@ -317,11 +316,7 @@ contract Swap is ISwap, Ownable {
   function cancel(uint256[] calldata nonces) external override {
     for (uint256 i = 0; i < nonces.length; i++) {
       uint256 nonce = nonces[i];
-      if (_markNonceAsUsed(msg.sender, nonce)) {
-        emit Cancel(nonce, msg.sender);
-      } else {
-        revert NonceAlreadyUsed(nonce);
-      }
+      _markNonceAsUsed(msg.sender, nonce);
     }
   }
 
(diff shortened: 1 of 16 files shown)


source/swap/contracts/Swap.sol
```diff
@@ -98,7 +98,7 @@ contract Swap is ISwap, Ownable {
   ) {
     if (_protocolFee >= FEE_DIVISOR) revert InvalidFee();
     if (_protocolFeeWallet == address(0)) revert InvalidFeeWallet();
-    uint256 currentChainId = getChainId();
+    uint256 currentChainId = block.chainid;
     DOMAIN_CHAIN_ID = currentChainId;
```
dmosites: Just combine these two lines and assign `DOMAIN_CHAIN_ID = block.chaind`?

source/swap/contracts/Swap.sol
```diff
@@ -495,8 +496,6 @@ contract Swap is ISwap, Ownable {
     address token,
     bytes4 kind
   ) internal {
-    // Ensure the transfer is not to self.
-    if (from == to) revert SelfTransferInvalid();
```
dmosites, line 485: Reason to remove this?

source/swap/contracts/Swap.sol
```diff
@@ -522,24 +517,21 @@ contract Swap is ISwap, Ownable {
...
   {
     uint256 groupKey = nonce / 256;
     uint256 indexInGroup = nonce % 256;
     uint256 group = _nonceGroups[signer][groupKey];
 
-    // If it is already used, return false
+    // If it is already used, return mit cancel and revert
     if ((group >> indexInGroup) & 1 == 1) {
-      return false;
+      emit Cancel(nonce, msg.sender);
```
dmosites, line 508: `Cancel` should only be emitted from the `cancel` function. Would place it after L319.

dmosites (changes requested):

source/swap/contracts/Swap.sol
smartcontrart: Makes ssenes

source/swap/contracts/Swap.sol
dmosites: Sorry, this was commented prior to the EIP712 commit.

source/swap/contracts/Swap.sol
dmosites: Happy to drop it and save gas.

source/swap/contracts/Swap.sol
```diff
@@ -509,28 +498,22 @@ contract Swap is ISwap, Ownable {
...
-    internal
-    returns (bool)
-  {
+  function _markNonceAsUsed(address signer, uint256 nonce) internal {
     uint256 groupKey = nonce / 256;
     uint256 indexInGroup = nonce % 256;
     uint256 group = _nonceGroups[signer][groupKey];
 
-    // If it is already used, return false
+    // If it is already used, return mit cancel and revert
```
dmosites, line 507: // If it is already used then revert

source/swap/contracts/Swap.sol
smartcontrart: Leaving as is then

source/swap/contracts/Swap.sol
smartcontrart: done

dmosites (approved):

PR #1001, opened 2023-01-25
Wrapper support for Swap and SwapERC20

dmosites:
Needs additional tests.


smartcontrart (approved):
Let's add tests when we revamp the test suite.

PR #1002, opened 2023-01-27
Changed nonce check from signatory to signer

smartcontrart:


source/swap/contracts/Swap.sol
```diff
@@ -495,21 +492,21 @@ contract Swap is ISwap, Ownable, EIP712 {
...
    */
-  function _markNonceAsUsed(address signer, uint256 nonce) internal {
+  function _markNonceAsUsed(address signatory, uint256 nonce) internal {
     uint256 groupKey = nonce / 256;
     uint256 indexInGroup = nonce % 256;
-    uint256 group = _nonceGroups[signer][groupKey];
+    uint256 group = _nonceGroups[signatory][groupKey];
 
-    // If it is already used, return mit cancel and revert
+    // If it is already used, return cancel and revert
```
dmosites, line 504: // Revert if nonce is already used

dmosites (changes requested):
Looks good, small comment change requested.

source/swap/contracts/Swap.sol
smartcontrart: agreed

dmosites (approved):
LGTM

PR #1005, opened 2023-01-30
TransferHandler to Adapter; Adapter[] provided in constructor

dmosites:
Also removes broken `integration.js` test file for Wrapper. To be replaced.


source/swap/contracts/Swap.sol
```diff
@@ -75,24 +74,28 @@ contract Swap is ISwap, Ownable, EIP712 {
...
 
-  // A registry storing a transfer handler for different token kinds
-  TransferHandlerRegistry public registry;
+  // Mapping of ERC165 interface ID to token Adapter
+  mapping(bytes4 => IAdapter) public adapters;
 
   /**
    * @notice Contract Constructor
    * @dev Sets domain for signature validation (EIP-712)
-   * @param swapRegistry TransferHandlerRegistry
```
smartcontrart, line 84: Should we replace the swapRegistry param by the _adapters param?

smartcontrart (approved):
looks good.

PR #1108, opened 2023-04-22
feat(Server): add get orders by filters and method

Mitsichury:


tools/libraries/src/Server.ts
```diff
@@ -20,9 +20,10 @@ import {
   ServerOptions,
   OrderResponse,
   SupportedProtocolInfo,
-  RequestFilter,
+  RequestFilterERC20,
   SortOrder,
   SortField,
+  RequestFilterMarketPlace,
```
dmosites, line 26: Should just be `RequestFilter` here.

tools/libraries/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@airswap/libraries",
-  "version": "4.0.5-beta.6",
+  "version": "4.0.5-beta.7",
```
dmosites, line 3: Also need to update the `@airswap/types` dep version below.

dmosites (changes requested):

tools/libraries/package.json
Mitsichury: done

tools/libraries/src/Server.ts
Mitsichury: done

dmosites (approved):

PR #1146, opened 2023-05-29
chore(server): Update response types following OrderFilter

Mitsichury:


tools/types/src/server.ts
```diff
@@ -22,28 +23,20 @@ export enum SortOrder {
...
   senderTokens?: string[]
+  senderWallet?: string
+  signerMaxAmount?: string
+  signerMinAmount?: string
+  signerTokens?: string[]
+  signerWallet?: string
   sortField?: SortField
   sortOrder?: SortOrder
-  offset: number
+  tokenIds?: string[]
```
dmosites, line 37: Probably want to split this into `signerIds` and `senderIds` as is done with the signer and sender tokens.

tools/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@airswap/types",
-  "version": "4.0.4",
+  "version": "4.1.0",
```
dmosites, line 3: We're still iterating on the minor version (and building the only clients using it so far) so let's just bump this to `4.0.5`.

tools/types/src/server.ts
Mitsichury: done

tools/types/package.json
Mitsichury: done

tools/types/src/server.ts
```diff
@@ -55,13 +49,10 @@ export type IndexedOrder<Type> = {
...
 }
 
 export type Pagination = {
-  first: string
-  last: string
-  prev?: string | undefined
-  next?: string | undefined
+  limit: number
+  offset: number
+  totalForQuery: number
```
piersss, line 57:
```suggestion
  total: number
```

I think Don said `total`. 😁

piersss (approved):
LGTM

PR #1147, opened 2023-06-05
Pool: simplify constructor, staking; remove mapping

dmosites:
Staking is only on mainnet so removed from constructor.
Combined staking token and contract into single setter.
Removed unused mapping from previous implementation.
Includes latest from develop.


smartcontrart (approved):
Looks good to me

PR #1148, opened 2023-06-05
registry refactor, terminology; add protocol cost

dmosites:
* Adds cost for protocol support (currently same as token support cost).
* Staking cost (formerly obligation cost) is now transferred during stake and unstake.
* Protocol and token support costs (formerly token cost) now transferred during add.
* Unstake removes server and all supported protocols and tokens.


wjmelements:
> Staking cost (formerly obligation cost) is now transferred during stake and unstake

This is an improvement to the design.

source/registry/contracts/Registry.sol
```diff
@@ -338,16 +230,135 @@ contract Registry {
...
+    }
+  }
+
   /**
    * @notice Return the staking balance of a given staker
    * @param _staker address of the account used to stake
    * @return balance of the staker account
    */
   function balanceOf(address _staker) external view returns (uint256) {
+    if (bytes(stakerServerURLs[_staker]).length == 0) return 0;
```
wjmelements, line 356: Interesting that you can:
* addTokens
* addProtocols

But your balance remains zero until you set the URL. Also, until you set the URL, you cannot removeStakedServer, but you can removeProtocols and removeTokens.
I think balanceOf should return the staking balance whether you have set the URL or not.

dmosites: Since they're not useful without a URL, functions `addTokens` and `addProtocols` could require a URL to be set. Thoughts?

source/registry/contracts/Registry.sol
```diff
@@ -16,190 +16,114 @@ contract Registry {
...
-  function setServerURL(string calldata _url) external {
+  function stakeForServer(string calldata _url) external {
+    if (bytes(_url).length == 0) revert ServerURLInvalid();
+
+    if (bytes(stakerServerURLs[msg.sender]).length == 0 && stakingCost > 0) {
+      stakingToken.safeTransferFrom(msg.sender, address(this), stakingCost);
+    }
     stakerServerURLs[msg.sender] = _url;
-    emit SetServerURL(msg.sender, _url);
+    emit StakeForServer(msg.sender, _url);
```
wjmelements, line 77: Is this a good event name for when we only update the URL?\

source/registry/contracts/Registry.sol
wjmelements: I want to use balanceOf to simplify my system so it's most important for me that it is accurate. Of secondary concern is the cost of updating the token list. Of minimal concern is the cost of balanceOf, since I don't think it will be invoked much on-chain.

smartcontrart (approved):

source/registry/contracts/Registry.sol
dmosites: How about `SetServer(sender, url)` and `UnsetServer(sender, url, protocols, tokens)`?

PR #1173, opened 2023-07-26
Implemented proposalId metric for root claims and a view function to …

smartcontrart:
…check claim by proposalId and user


source/pool/contracts/Pool.sol
```diff
@@ -128,34 +131,55 @@ contract Pool is IPool, Ownable2Step {
...
-    if (roots[_root] == false) {
-      roots[_root] = true;
+    if (roots[rootIds[_proposalId]] == false) {
+      roots[rootIds[_proposalId]] = true;
     }
     for (uint256 i = 0; i < _accounts.length; i++) {
       address account = _accounts[i];
-      if (claimed[_root][account]) revert AlreadyClaimed();
-      claimed[_root][account] = true;
+      if (claimed[rootIds[_proposalId]][account]) revert AlreadyClaimed();
```
gpxl-dev, line 146: This could cause an issue - we switched to `proposalId` instead of root to avoid the case where two proposals share the same `root` (the same voters, each with the same points - this will be more common than it sounds because sometimes there are multiple proposals for the same snapshotted block).

gpxl-dev (changes requested):
Suggest removing current`roots` mapping completely.

Would rename `rootIds` to `rootsByProposalId` (or even `roots` naming up to you, but this neither the key nor value is a `rootId`).

In order to know if claims are enabled for a proposal you just need to check that the root isn't 0.

source/pool/contracts/Pool.sol
```diff
@@ -128,34 +131,55 @@ contract Pool is IPool, Ownable2Step {
...
-    emit Enable(_root);
+    emit Enable(rootIds[_proposalId]);
   }
 
   /**
    * @notice Enables claims for a merkle tree of a set of scores
    * @param _root bytes32
    */
-  function enable(bytes32 _root) external override multiAdmin {
+  function enable(
```
gpxl-dev, line 156: Really minor, but probably worth switching the order of these now, `enable(proposalId, root)` - "enable the proposal by adding this root".

smartcontrart: Done in the new commit

source/pool/contracts/Pool.sol
smartcontrart: Done in the new commit

source/pool/contracts/Pool.sol
```diff
@@ -128,34 +128,53 @@ contract Pool is IPool, Ownable2Step {
...
+    // }
     for (uint256 i = 0; i < _accounts.length; i++) {
       address account = _accounts[i];
-      if (claimed[_root][account]) revert AlreadyClaimed();
-      claimed[_root][account] = true;
+      if (claimed[rootsByGroupId[_groupId]][account]) revert AlreadyClaimed();
+      claimed[rootsByGroupId[_groupId]][account] = true;
     }
-    emit Enable(_root);
+    emit Enable(rootsByGroupId[_groupId]);
```
gpxl-dev, line 146: Setting as claimed shouldn't emit an enable.

source/pool/contracts/Pool.sol
```diff
@@ -25,15 +25,15 @@ contract Pool is IPool, Ownable2Step {
   // Max percentage for a claim with infinite score
   uint256 public max;
 
-  // Mapping of tree root to boolean to enable claims
-  mapping(bytes32 => bool) public roots;
-
   // Mapping of address to boolean to enable admin accounts
   mapping(address => bool) public admins;
 
   // Mapping of tree root to account to mark as claimed
```
gpxl-dev, line 31: Whilst we have now massively reduced the possibility of two roots that are the same, it could still happen (no change to voters or token holdings between votes).
The key here should be the `groupId`

source/pool/contracts/Pool.sol
```diff
@@ -128,34 +128,53 @@ contract Pool is IPool, Ownable2Step {
...
-    bytes32 _root,
+    bytes32 _groupId,
     address[] memory _accounts
   ) external override multiAdmin {
-    if (roots[_root] == false) {
-      roots[_root] = true;
-    }
+    // if (roots[rootsByGroupId[_groupId]] == false) {
+    //   roots[rootsByGroupId[_groupId]] = true;
+    // }
```
gpxl-dev, line 140: Forgot to remove this :)

source/pool/contracts/Pool.sol
```diff
@@ -128,34 +128,53 @@ contract Pool is IPool, Ownable2Step {
...
     for (uint256 i = 0; i < _accounts.length; i++) {
```
gpxl-dev, line 141: You will save a lot of gas here if you extract `bytes32 root = rootsByGroupId[_group]` outside this for loop. As it is you're reading from storage on every iteration for something that isn't going to change.

Even without the loop, you're reading this from storage twice, once here, and once at the `Emit` line lower down. 

**NOTE** if you agree that `claimed` should be keyed by the `groupId` then you will need to change this anyway so the comment may not be relevant.

source/pool/contracts/Pool.sol
```diff
@@ -128,34 +128,53 @@ contract Pool is IPool, Ownable2Step {
...
    * @param _root bytes32
    */
-  function enable(bytes32 _root) external override multiAdmin {
-    if (roots[_root]) revert RootExists(_root);
-    roots[_root] = true;
+  function enable(
+    bytes32 _groupId,
+    bytes32 _root
+  ) external override multiAdmin {
+    if (rootsByGroupId[_groupId] != 0) revert GroupIdExists(_groupId);
```
gpxl-dev, line 157: Question for @dmosites whether setting the root should be permanent. 
My feeling is that it should be possible to overwrite a root, because it's something that's quite hard to verify when the transaction is submitted, so I can see errors occurring with this...

source/pool/contracts/Pool.sol
```diff
@@ -128,34 +128,53 @@ contract Pool is IPool, Ownable2Step {
...
+    rootsByGroupId[_groupId] = _root;
     emit Enable(_root);
```
gpxl-dev, line 159: I would personally emit both `groupId` and `root` here, but if I was going to choose only  one it would be `groupId`.

source/pool/contracts/Pool.sol
```diff
@@ -257,13 +276,21 @@ contract Pool is IPool, Ownable2Step {
     Claim memory _claim;
     for (uint256 i = 0; i < _claims.length; i++) {
```
gpxl-dev, line 277: Store the root in memory for the duration of each loop to avoid 4 or 5 extra reads from storage and save gas.

`bytes32 root = rootsByGroupId[_claim.groupId];`

gpxl-dev (changes requested):
I think `claimed` should be keyed by a `groupId`, not a `root`. 
Also a couple of other small changes.

gpxl-dev (changes requested):
I think `claimed` should be keyed by a `groupId`, not a `root`. 
Also a couple of other small changes.

source/pool/contracts/Pool.sol
smartcontrart: indeed thanks for the catch

source/pool/contracts/Pool.sol
```diff
@@ -128,32 +128,48 @@ contract Pool is IPool, Ownable2Step {
   /**
    * @notice Set claims from previous pool contract
    * @dev Only owner
-   * @param _root bytes32
+   * @param _groupId bytes32
    * @param _accounts address[]
    */
   function setClaimed(
```
dmosites, line 134: This function can be removed.

source/pool/contracts/Pool.sol
```diff
@@ -128,32 +128,48 @@ contract Pool is IPool, Ownable2Step {
...
+    rootsByGroupId[_groupId] = _root;
+    emit Enable(_groupId, _root);
+  }
+
+  /**
+   * @notice Returns the claim status of a root for a given address
+   * @param _address address
+   * @param _groupIds bytes32[]
+   */
+  function hasClaimedGroups(
```
dmosites, line 164: How about `getClaimStatuses`?

source/pool/contracts/Pool.sol
```diff
@@ -128,46 +128,42 @@ contract Pool is IPool, Ownable2Step {
...
   }
 
   /**
    * @notice Returns the claim status of a root for a given address
    * @param _address address
-   * @param _groupIds bytes32[]
+   * @param _trees bytes32[]
    */
-  function hasClaimedGroups(
+  function hasClaimedTrees(
```
dmosites, line 142: Something like `getClaimStatusForTrees`? Similar to Registry.sol.

source/pool/contracts/Pool.sol
```diff
@@ -128,46 +128,42 @@ contract Pool is IPool, Ownable2Step {
...
     address _address,
```
dmosites, line 143: `address _account,`

dmosites (approved):

source/pool/contracts/Pool.sol
```diff
@@ -252,37 +246,43 @@ contract Pool is IPool, Ownable2Step {
...
       _claim = _claims[i];
-      if (!roots[_claim.root]) revert RootDisabled(_claim.root);
-      if (claimed[_claim.root][msg.sender]) revert AlreadyClaimed();
-      if (!verify(msg.sender, _claim.root, _claim.score, _claim.proof))
-        revert ProofInvalid(_claim.root);
-      _totalScore = _totalScore + _claim.score;
-      claimed[_claim.root][msg.sender] = true;
-      _rootList[i] = _claim.root;
+      if (rootsByTree[_claim.tree] == 0) revert TreeDisabled(_claim.tree);
+      if (claimed[_claim.tree][msg.sender]) revert AlreadyClaimed();
```
gpxl-dev, line 255: As previously mentioned it saves gas in this function to store `rootsByTree[_claim.tree]` in an intermediate variable at the start of this loop as we check it several times..
Any gas savings are important in this function for small stakers.

@dmosites also mentioned iterating in reverse and linked [this resource](https://hackmd.io/@totomanov/gas-optimization-loops).

source/pool/contracts/Pool.sol
```diff
@@ -252,37 +246,43 @@ contract Pool is IPool, Ownable2Step {
     uint256 _minimumAmount
   ) internal returns (uint256, bytes32[] memory) {
     if (_claims.length <= 0) revert ClaimsNotProvided();
-    uint256 _totalScore = 0;
+    uint256 _totalValue = 0;
     bytes32[] memory _rootList = new bytes32[](_claims.length);
```
gpxl-dev, line 250: This should be `treeList` as we are more interested in which trees were claimed than what their roots were.

source/pool/contracts/Pool.sol
```diff
@@ -252,37 +246,43 @@ contract Pool is IPool, Ownable2Step {
...
+      if (rootsByTree[_claim.tree] == 0) revert TreeDisabled(_claim.tree);
+      if (claimed[_claim.tree][msg.sender]) revert AlreadyClaimed();
+      if (
+        !verify(
+          msg.sender,
+          rootsByTree[_claim.tree],
+          _claim.value,
+          _claim.proof
+        )
+      ) revert ProofInvalid(rootsByTree[_claim.tree]);
```
gpxl-dev, line 263: Similar to above, the event should be emitting a tree ID not a root.

gpxl-dev (changes requested):
Small changes to event emission and a gas optimisation suggestion.

gpxl-dev (changes requested):
Small changes to event emission and a gas optimisation suggestion.

dmosites (approved):

gpxl-dev (approved):

PR #1174, opened 2023-07-28
Registry integration tests

smartcontrart:


source/registry/package.json
```diff
@@ -20,7 +20,6 @@
     "typechain": "tsc -b",
     "coverage": "hardhat coverage",
     "test": "hardhat test",
-    "test:unit": "hardhat test ./test/unit.js",
```
dmosites, line 23: `test:unit` is called by CI so let's just update the file name here to test/Registry.js.

smartcontrart: That won't run test/RegistryIntegration.js. Is that ok?

dmosites: Good point, updated here: https://github.com/airswap/airswap-protocols/pull/1176

smartcontrart: @dmosites I believe this is not resolved. Correct? `"test": "hardhat test",` will run both Registry.js and RegistryIntegration.js.

Let me know if there's anything else to be done on that.

dmosites (approved):

PR #1224, opened 2023-10-22
remove redundant list

Mitsichury:


dmosites:
Thanks @Mitsichury. This is by design (`params` is a list of arguments, the only argument is `pairs`). Closing as `getPricingERC20` is to be updated shortly with named params anyway.

PR #1225, opened 2023-10-23
named param for getPricingERC20; add minExpiry; response validation

dmosites:


tools/libraries/src/Server.ts
```diff
@@ -478,7 +520,8 @@ export class Server extends TypedEmitter<ServerEvents> {
       if (
         typeof flat === 'object' &&
         param in flat &&
-        flat[param].toLowerCase() !== params[param].toLowerCase()
+        String(flat[param]).toLowerCase() !==
```
Mitsichury, line 523: This cast on an object can end up with this representation : '[object Object]' for instance if flat[param] or params[param] is an object: [[]] or [{}] etc...

tools/libraries/src/Server.ts
```diff
@@ -220,21 +239,38 @@ export class Server extends TypedEmitter<ServerEvents> {
...
     })
   }
 
   public async getPricingERC20(
-    pairs: { baseToken: string; quoteToken: string }[]
+    pairs: { baseToken: string; quoteToken: string }[],
+    minExpiry?: string
   ): Promise<Pricing[]> {
-    return this.callRPCMethod<Pricing[]>('getPricingERC20', [pairs])
+    return this.callRPCMethod<Pricing[]>('getPricingERC20', {
```
Mitsichury, line 270: Should we also check the fields in the body here ?

tools/libraries/src/Server.ts
Mitsichury: maybe somehting like that ? 

  
  private compare(params: any, flat: any): Array<string> {
    const errors: Array<string> = []
    for (const param in params) {
      if (typeof flat === 'object' && param in flat ) {
        const flatP= flat[param]
        const paramP= params[param]
        if(typeof flatP != typeof paramP) {
          errors.push(param)
        }
        else if(typeof flatP == 'object') {
          if (compare(flatP, paramP)) {
            errors.push(param)
          }
        }
        else if(String(flatP).toLowerCase() !== String(paramP).toLowerCase()) {
          errors.push(param)
        }
      }else {
        errors.push(param)
      }
    }
    return errors
  }

dmosites: Request and response params for `get*OrderERC20` calls are shallow so these will only ever be stringish. Since it's only being used in those calls it should be okay.

tools/libraries/src/Server.ts
dmosites: Fields are checked on the `get*OrderERC20` calls because you want the response params to match the request params. Since the `getPricingERC20` response is more complex (with lists of pairs and lists of levels) I say we just let it go through to the client and they can work it out.

Mitsichury: Okay !

Mitsichury (approved):

PR #1306, opened 2024-04-25
Use EIP712s hashTypedData functions

k06a:


dmosites:
Thanks. This could also be used on SwapERC20.sol:L260?
https://github.com/airswap/airswap-protocols/pull/1306/files#diff-a1ad63b6dfc142aa0e9d5894dd4d55c81fadc2bd1a98318b964be3c84c91aae7L260

k06a:
@dmosites true

dmosites:
Implemented in #1318 thank you @k06a

PR #1320, opened 2024-07-25
Delegates

smartcontrart:


source/delegate/contracts/Delegate.sol
```diff
@@ -106,12 +110,15 @@ contract Delegate is IDelegate, Ownable {
...
 
     if (
-      _senderAmount > (_signerAmount * rule.senderAmount) / rule.signerAmount
+      _signerAmount <
+      (rule.signerAmount * (rule.senderRuleAmount - rule.senderFilledAmount)) /
+        rule.senderRuleAmount
     ) {
       revert InvalidSignerAmount();
     }
+    if (rule.ruleExpiry < _expiry) revert RuleExpired();
```
dmosites, line 119: This should compare to block timestamp rather than order expiry:
```
if (rule.ruleExpiry <= block.timestamp) revert RuleExpired();
```

dmosites (approved):

Issue #1316, opened 2024-07-05
Add deploys-commits.js to contract packages


dmosites:
To track the code with which a each package version was deployed.

PR #1321, opened 2024-08-06
Scripts enhancements

smartcontrart:
Solves issue #1316

### DIFF
--- a/source/batch-call/deploys-commits.js
+++ b/source/batch-call/deploys-commits.js
@@ -1,0 +1,1 @@
+module.exports = {}
--- a/source/batch-call/deploys-commits.js.d.ts
+++ b/source/batch-call/deploys-commits.js.d.ts
@@ -1,0 +1,1 @@
+declare module '@airswap/batch-call/deploys-commits.js'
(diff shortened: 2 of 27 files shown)


populate_deploy_commits.js
```diff
@@ -0,0 +1,83 @@
...
+  'swap-erc20',
+  'wrapper',
+]
+
+function getLatestCommitForContract(contractPath) {
+  const branches = ['main', 'develop']
+  for (const branch of branches) {
+    try {
+      const commit = execSync(
+        `git log -1 --format=%H ${branch} -- ${contractPath}/contracts`
```
dmosites, line 21: Let's check `${contractPath}/contracts/deploys.js` specifically.

dmosites:
Because `populate_deploy_commits.js` is used once it might not need to be tracked. If you think it should be tracked please move to `scripts/populate-commits.js` or similar and include a file comment describing its use.

smartcontrart:
Agreed - `populate_deploy_commits.js` shouldn't be tracked - I added just so you could have a look at the logic and see if there's anything that didn't make sense

dmosites (approved):

PR #1352, opened 2025-02-05
Fix rounding management of partial fills

smartcontrart:


source/delegate/test/Delegate.js
```diff
@@ -412,6 +412,115 @@ describe('Delegate Unit', () => {
       ).to.emit(delegate, 'DelegatedSwapFor')
     })
 
+    it('successfully swaps with a rounding error - Upper bound', async () => {
```
dmosites, line 415: How about "successfully swaps with rounded-down values"? Also, if not already, it may be worth noting in the Delegate documentation that all amounts are "atomic" units (e.g. Wei).

dmosites (approved):

PR #1355, opened 2025-03-06
Created helper function to calculate Delegate SignerAmount

smartcontrart:


makkelie-dev (changes requested):
This will only work if the sender and signer token have the same decimals. If the decimals are different then the ratio is off by a few magnitudes.

After you fixed this add a few tests to demonstrate that this works.

makkelie-dev (approved):
LGTM

PR #1357, opened 2025-03-17
Created getWalletNFTBalances and getWalletNFTAllowances functions and…

smartcontrart:
… related tests


source/batch-call/contracts/BatchCall.sol
```diff
@@ -14,6 +16,7 @@ contract BatchCall {
   using Address for address;
 
   error ArgumentInvalid();
+  error UnsupportedTokenType();
```
dmosites, line 19: Unused

dmosites (approved):

